PR #188, opened 2014-10-29
fixed problem with special umlauts in emos adapter

steveh80:
Old code didn’t work with special umlauts like in “BOÎTE À DÉSIR”

Detailed explanation:
The code wants to remove non breakable spaces. By replacing chr(0xa0) - which is nbsp for single-byte character sets - it destroys characters in multibyte character sets that are using that character for a multibyte character like À (which is composed by 0xc3 AND 0xa0).


stasiukaitis-saulius:
Didn't receive an OCA yet, have to close it.

PR #191, opened 2014-12-08
fixing current url matching in oxUtilsServer

robinlehrmann:
The URL doesn't match if ```$this->getServerVar('SCRIPT_NAME')``` is set.

I have created a module and overwrite the "oxDynImgGenerator". Then I change the .htaccess like this:
```
RewriteCond %{REQUEST_URI} (\/out\/pictures\/generated\/)
RewriteCond %{REQUEST_FILENAME} !-f
RewriteCond %{REQUEST_FILENAME} !-d
RewriteRule (\.jpe?g|\.gif|\.png|\.svg)$ modules/example/getimg.php
```

In the enterprise version matches the oxbaseshop with shopid "1" everytime because the condition never matches.

So I get a redirect on the "mallshopstart" template.

I think it is a highly privileg hotfix and you should merge it in this version.

Cheers, Robin

### DIFF
--- a/source/core/oxutilsserver.php
+++ b/source/core/oxutilsserver.php
@@ -443,7 +443,7 @@ class oxUtilsServer extends oxSuperCfg
         $sCurrentHost = str_replace('/', '', $sCurrentHost);
         $sURL = str_replace('/', '', $sURL);
 
-        if ($sURL && $sCurrentHost && strpos($sURL, $sCurrentHost) !== false) {
+        if ($sURL && $sCurrentHost && strpos($sURL, $sServerHost) !== false) {
             //bug fix #0002991
             if ($sUrlHost == $sRealHost) {
                 return true;


robinlehrmann:
Hi,

It should be a part of the shoplist. Create a new method in oxShopList to get the Shop by url.
oxShopList->loadShopByUrl($sURL); (-> load oxShopUrlList, iterate it and return a array with shop instance)

To check if the current url matches with the url, move the method "isCurrentUrl" to the oxShop-model.
oxShop->matchUrl($sURL);

Additionally I would move the MallShopUrls and the LanguageUrls in a seperate table with One2Many Relation:
e.g.:
oxShopUrls:
OXID
OXSHOPID
OXPROTOCOL
OXPATH
OXLANG
OXURL

With this table you could create a list "oxShopUrlList".

In this list is a method like "loadByUrl($sUrl)". (In this method the URL should splitted by protocol, url and path. The protocol and the url should be equal but the path should match, if strpos is match. A good idea is to create a oxUrl-Model -> e.g. like Zend: http://framework.zend.com/manual/current/en/modules/zend.uri.html)

With this solution you have the urls in one table and you can create optional multiple shopUrls for one shop.

stasiukaitis-saulius:
Let me think a few days, have to finish other stuff first. I'll try to answer in a few days, hang on.

robinlehrmann:
any news?

artsev-oxid:
Hi,
Hopefully, I can answer you during this week.
Also, please sign the OCA using our new service (https://www.clahub.com/agreements/OXID-eSales/oxideshop_ce)

artsev-oxid:
Hi Robin,

Let's see if I get your suggestion right: you suggest to have an shop list class which would return the shop by it's URL and a language list class which would return the language by it's URL. You also provide several implementation tips on how to achieve that.

At this point I would like to skip details and first off point out that both of those solutions basically already exist in OXID eShop: oxUtilsObject works with shop list and oxLang works with language list. Both, oxUtilsObject as well as oxLang, check if the currently requested URL matches one of the list entries. Also, language and shop ID checks are applied in the oxUtilsServer method isCurrentUrl.

The structure where a different shop ID might be separated not only by host but also by it's path is already supported. That's why “SCRIPT_NAME” is added to the host.

Changing implementation details to one way or another would not solve your problem in general.

As I said before: the shop might be separated by URL, but in this case we have to check against the current URL and not only against the current domain.

What you actually did in your pull request is to remove the URL check in all cases against the current URL and left only the check against the current domain. But this will break the existing solution.

Unfortunately, I can't merge your pull request for this reason.

robinlehrmann:
Why are these checks located in oxUtilsObject and oxUtilsServer? Why should these classes know anything about Shop handling?

The OXID-Framework must load the complete config to check which url is mapped to an URL. This is over powered e.g. for a small script.

The reason to refactor the language handling:

e.g.

We have a module called vkf/meta

This module should add hreflang tags for products.
hreflang tags are for SEO to prevent duplicate content.

(https://support.google.com/webmasters/answer/189077?utm_source=wnc_108502&utm_term=link_5&utm_content=uns_3d3809f5a8000000&utm_campaign=t_1382697016647190&utm_medium=mc)

With my suggestion you can build a nice SQL-Query with oxseo, oxarticles and oxshopurls table.

However..

You can add a "str_replace" in the new method "matchUrl":

public function matchUrl($sUrl)
{
    // trims the unnecessary path of url
    $sPathToRemove = str_replace($this->getMallShopUrl()->getFullUrl(), '', $sUrl);
    $sUrlToCheck       = str_replace($sPathToRemove, '', $sUrl);
    return ($sUrlToCheck === $this->getMallShopUrl()->getFullUrl());
}

Why did you close this pull-request? I don't understand why close everthing premature!?

artsev-oxid:
Dear @robinlehrmann,

Your suggestions about the refactoring of the language handling are absolutely worth to be discussed, but unfortunately misplaced in this pull request.
As I explained above, we can't merge your committed code because it would break the present concept.

If you want to talk about the concept itself, the best way is most likely to use dev-general mailing list. Please do not forget to deliver a good business case for it so everybody can clearly understand what you want to achieve. Once this domain is clear, we might talk about how to achieve this behaviour from the architecture and code point of view.

robinlehrmann:
Okay, I will write to the mailing.

What do you thing about my suggestion, to trim the unnecessary path of url with the shop url?

e.g.:

public function matchUrl($sUrl)
{
// trims the unnecessary path of url
$sPathToRemove = str_replace($this->getMallShopUrl()->getFullUrl(), '', $sUrl);
$sUrlToCheck = str_replace($sPathToRemove, '', $sUrl);
return ($sUrlToCheck === $this->getMallShopUrl()->getFullUrl());
}

artsev-oxid:
Hi,
Sorry, I can't tell you how good this looks for me, because first of all I should be able to see where and when this new method will be used.
I suggest you to offer this suggestion in a separate pull request together with added PHPUnit tests :ok_hand:
Also beware that existing tests probably should be adapted along the way :exclamation:

PR #196, opened 2014-12-15
Write review without login with userhash

suabo:
To allow user write a review without login you need to assign the hash earlyer in Template and check it like oxcmp_user.


kermie:
@suabo, we'd like to merge your PR. Still waiting for your signed OCA.

kermie:
Received the signed OCA, everything seems to be alr8 now. Thanks!

PR #197, opened 2015-01-04
add full functional Travis CI configuration

adriankirchner:
When looking through previous pull requests you may notice that very few of them ship with unittests. I think one reason for this is the lack of configuration documentation about the environment you need to successfully execute the whole OXID eShop testsuite and another reason may be the effort you need to invest to configure the required software stack (with PHP version matrix) including all dependencies.

I'm aware of the repository tomasliubinas/vagrant-oxideshop but this requires the contributor to install a bunch of additional software packages besides the fact that the chef recipes don't work with current versions of chef and Sebastian Bergmann did shutdown the corresponding pear channel. Testsuite execution over different versions of PHP is also not possible with this solution.

I'm aware of the pull request OXID-eSales/oxideshop_ce#148 and the fact that you rejected it. In contrast to the referenced pull request the Travis CI configuration in my pull request is fully functional so the whole latin1 testsuite is executed across all four supported PHP versions without errors (see [1]). Furthermore you pointed out:
> [..] Have the feeling that we have to do that from scratch, a modular way, using our internal infrastructure and put much more effort in it than we could do now (as the focus is presently somewhere else) [..]

I respect your dedication to do things from scratch and in a modular way. But for us external contributors, without access to your internal infrastructure, Travis CI is the perfect solution with nearly no overhead and direct availability not just in the (maybe far) future.
Simply don't see Travis CI as a part of your internal test and release process but as a nice and quick feedback for external contributors as well as a first check for incoming pull requests.
I separated the very minor changes to the existing codebase into different commits so you can quickly determine that there is no impact of any execution outside of the Travis CI environment. Knowing that, you will see that there is nothing to do for you if you merge this pull request.

With the need of signing your OCA (which is inevitable) before contributing, there is already a barrier which is uncommon in the opensource world (at GitHub). It should be in your interest to lower contribution barriers.

I expect this setup to be as stable as a setup with that much software and configuration dependencies could be. However, should any problems occur in the future or the development of OXID-eSales/testing_library results in necessary modifications, feel free to ping me and I'll have a look into it.

After merging this pull request, you only need to sign in at Travis CI with a GitHub account that has access to this repository. Remember to enable the checkbox `Build only if .travis.yml is present` in the project settings. 

[1] https://travis-ci.org/adriankirchner/oxideshop_ce

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,52 @@
+language: php
+
+notifications:
+  email: false
+
+php:
+  - "5.3"
+  - "5.4"
+  - "5.5"
+  - "5.6"
+
+# TRAVIS_ERROR_LEVEL = E_ALL & ~E_DEPRECATED & ~E_NOTICE & ~E_STRICT
+env: TRAVIS_ERROR_LEVEL=22519 oxINSTALLSHOP=0 oxSKIPSHOPSETUP=1 OXID_VERSION=CE OXID_EDITION=CE oxPATH="$TRAVIS_BUILD_DIR/source/"
+
+services: mysql
+
+before_install:
+  - echo 'Europe/Berlin' | sudo tee /etc/timezone
+  - sudo dpkg-reconfigure --frontend noninteractive tzdata
+
+before_script:
+  # apache setup (http://docs.travis-ci.com/user/languages/php/#Apache-%2B-PHP)
+  - sudo apt-get install apache2 libapache2-mod-fastcgi
+  # enable php-fpm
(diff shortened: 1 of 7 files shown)


kermie:
Thanks @adriankirchner, this is amazing! :-)
Still waiting for some peeps to come back from the holidays, hang on a second.

adriankirchner:
alright ;)

kermie:
Reviewing your contrib within the next sprint.

PR #208, opened 2015-01-25
Change getModuleUrl to return HTTPS hyperlink

bertrandjackermann:
Enables the method to return an HTTPS hyperlink for usage in admin modules. Important when the shop admin is configured to work with HTTPS.

### DIFF
--- a/source/core/oxviewconfig.php
+++ b/source/core/oxviewconfig.php
@@ -1362,6 +1362,10 @@ class oxViewConfig extends oxSuperCfg
             rtrim($this->getConfig()->getCurrentShopUrl(false), '/'),
             $this->getModulePath($sModule, $sFile)
         );
+        
+		if (($this->getConfig()->getConfigParam('sAdminSSLURL') != null) && ($this->isAdmin())){
+			$sUrl = str_replace('http:','https:',$sUrl);
+		}
 
         return $sUrl;
     }


ooxi:
Are you sure that works? HTTP and HTTPS URLs can be totally different

alfredbez:
The following code only replaces the protocol:
```php
$sUrl = (strtolower(substr($sUrl, 0, 5)) === 'http:') ? 'https:' . substr($sUrl, 5) : $sUrl;
```

bertrandjackermann:
Hello, yes, this only adds an "s". But the if statement checks whether we are running the admin and also the sslAdminURL. So it makes sense.

SvenBrunk:
Or maybe 
```php
$sURL = preg_replace('/^http:/','https:', $sURL);
```

ooxi:
The main argument against this patch was not that changing the scheme does not work but that it's not the correct thing to do!

tomasliubinas:
Please note:
http://bugs.oxid-esales.com/view.php?id=5911

MantasVaitkunas:
Hi, need some more time to check it, hang on for a couple of more days.

bertrandjackermann:
@ooxi Yes, it only changes the protocol -- but it changes the shop's address when the admin is used via HTTPS, so css files and other can be loaded.

MantasVaitkunas:
@bertrandjackermann, did you check @tomasliubinas comment? In the reported bug there is a note which suggests to use config option **sSSLShopURL**. This option is used in **getModuleUrl** method when shop is accessed via https.

Sieg:
Hello @bertrandjackermann, 
Please head back when you want to discuss about this PR again. Closing it for now.

PR #209, opened 2015-01-30
Adding checkbox for configurable performance check in dashboard

robinlehrmann:
We have 17 languages and 32 shops in our Shop. If we login to admin panel, the admin checks the collations on nearly every klick. This makes a high workload for the backend and the backend is very sluggish.

I think it would be a good idea to add a configurable checkbox in performance settings for systemrequirement check in dashboard.

### DIFF
--- a/source/application/controllers/admin/navigation.php
+++ b/source/application/controllers/admin/navigation.php
@@ -204,9 +204,15 @@ class Navigation extends oxAdminView
     { // #661
         $aMessage = array();
 
-        // check if system reguirements are ok
-        $oSysReq = new oxSysRequirements();
-        if (!$oSysReq->getSysReqStatus()) {
+        if ($this->getConfig()->getConfigParam('blCheckSysReq')) {
+            // check if system reguirements are ok
+            $oSysReq = new oxSysRequirements();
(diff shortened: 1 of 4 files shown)


ooxi:
:+1:

artsev-oxid:
Hi,

I merged your enhancement. But did some changes to it considering these facts:

- blCheckSysReq must be true by default

- System requirement check MUST be performed when accessing admin dashboard BY DEFAULT (blCheckSysReq is set to TRUE or NOT SET). A correct message must be displayed (NAVIGATION_SYSREQ_MESSAGE) with a type of warning.  

- System requirement check MUST NOT be performed when accessing admin dashboard when blCheckSysReq is set to FALSE. A correct message must be displayed (NAVIGATION_SYSREQ_MESSAGE_INACTIVE) with a type of message.

tomasliubinas:
Also please check the fix by @Buratinas which was intended to completely eliminate this system requirement check performance problem for backend. Afaik the fix was tested with tens of thousands of db views causing no performance penalty: bc858d1886c2f53a9bb8f4d0f03949494e206504

PR #210, opened 2015-01-30
Add Fallback Value to getParam functions

revoltek-daniel:
Add an Fallback Param to getConfigParam and getRequestParameter to ensure a value is available also if the var could not load.

### DIFF
--- a/source/core/oxconfig.php
+++ b/source/core/oxconfig.php
@@ -310,12 +310,12 @@ class oxConfig extends oxSuperCfg
     /**
      * Returns config parameter value if such parameter exists
      *
-     * @param string $sName config parameter name
-     * @param mixed $sFallback fallback value if no config var is found default null
+     * @param string $sName    config parameter name
+     * @param mixed  $sDefault default value if no config var is found default null
      *
      * @return mixed
      */
-    public function getConfigParam($sName, $sFallback = null)
+    public function getConfigParam($sName, $sDefault = null)
     {
         if (defined('OXID_PHP_UNIT')) {
             if (isset(modConfig::$unitMOD) && is_object(modConfig::$unitMOD)) {
@@ -329,14 +329,14 @@ class oxConfig extends oxSuperCfg
         $this->init();
 
         if (isset ($this->_aConfigParams[$sName])) {
-            return $this->_aConfigParams[$sName];
+            $sValue = $this->_aConfigParams[$sName];
+        } elseif (isset($this->$sName)) {
+            $sValue = $this->$sName;
+        } else {
+            $sValue = $sDefault;
         }
 
-        if (isset($this->$sName)) {
-            return $this->$sName;
-        }
-        
-        return $sFallback;
+        return $sValue;
     }
 
     /**
@@ -634,11 +634,10 @@ class oxConfig extends oxSuperCfg
      *
      * @param string $sName     Name of parameter
      * @param bool   $blRaw     Get unescaped parameter
-     * @param strin  $sFallback fallback value if no config var is found default null
      *
      * @return mixed
      */
-    public function getRequestParameter($sName, $blRaw = false, $sFallback = null)
+    public function getRequestParameter($sName, $blRaw = false)
     {
         if (defined('OXID_PHP_UNIT')) {
             if (isset(modConfig::$unitMOD) && is_object(modConfig::$unitMOD)) {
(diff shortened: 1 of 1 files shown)


robinlehrmann:
You can check the value with is_null. Why do you need a Fallback?

revoltek-daniel:
!== will be a little faster than !is_null. 

I like this Syntax more than the use of this funktion.

Sometimes you need a fallback if the value is not provided in the Request.

```
$iPage = $oConfig->getRequestParameter('page');
if (!$iPage) {
     $iPage = 0;
}
```

i like this version more:
```
$iPage = $oConfig->getRequestParameter('page', false, 0);
```

robinlehrmann:
Okay. In your example you can cast to integer:

$iPage = (int) $oConfig->getRequestParameter('page');

But I think a fallback is a good idea for example:

$sListType = $oConfig->getRequestParameter('listtype');
if (!$sListType) {
     $sListType = 'list';
}

$sListType = $oConfig->getRequestParameter('listtype', false, 'list');

jkrug:
Please don't merge this! It makes code simply less understandable.

revoltek-daniel:
@jkrug why would this make code less understandable?

Every Major Framework (Zend, Symfony, Laravel, Yii) use this behavior? 
You could use it but you musst not, you could use the old behavior, but you have the option to rely that the function return the defined value.

MantasVaitkunas:
Hi, I would say idea is good. But some minor changes could be done:
$sFallback parameter could be changed to $sDefault and unit test could be provided.

tabsl:
why good idea? if someone would like to check this, he can do this with his own if statement. so i can look up all oxid functions like this and add this fallback. good idea, or? i thought shop core must be slimmer?

MantasVaitkunas:
@tabsl, I completely agree that shop core must be slimmer, but if this PR will make developers life easier, why not? Maybe some code parts of these methods could be extracted to reduce complexity.
And talking about additional parameter in methods, I just see it useful as sometimes request parameter could not be found, so I would like to return default value.

stasiukaitis-saulius:
Hi everyone,

Having a fallback in general makes code less readable as one can not see which arguments are meaningful and which one is fallback. It also almost adds no value as this can be easily reached with a condition statement. However getting request parameter might be an exception. @revoltek-daniel already mentioned that other frameworks use this so it might be common sense to expect second parameter as fallback in request class getter. Another argument for this solution would be that Symfony use this so it's quite likely that OXID eShop will use this structure after some time anyway.
Another argument against this solution is that it can also lead to code duplication. For example if you have “list” as default value for “listtype” it might be better to have getter for “listtype”. This getter would call request object or would return default value if “listtype” is not set in request. Having clever getter in request handling class would encourage to bypass such a solution.

In general for me it is like 50% - 50% to accept or reject this pull request.

Buratinas:
Hi,

We had an internal discussion about this pull request and how to integrate this into the core. We think that it would be best to have default parameters as a second option, the way it is in `getConfigParameter`. However `getRequestParameter` method already has a second parameter, which makes the usage a bit uncomfortable when you need to use it:
`$sListType = $oConfig->getRequestParameter('listtype', false, 'list');`

We came to a decision how to possibly move forward with this, and what steps could be taken to get there. We thought about implementing a new method, called `getRequestRawParameter`, which by default would return raw value of request. As escaping is not a part of input validation and more about output validation, this method seems logical. Some information about filter input/escape output principle: [wiki](http://en.wikipedia.org/wiki/Secure_input_and_output_handling), [other source](http://terrychay.com/article/php-advent-security-filter-input-escape-output.shtml).

As for what could be improved in order for us to accept the pull request, here's the list:

1. Rename `$sFallback` to `$sDefault`.
1. Write unit tests.
1. `getConfigParameter` has 2 return statements. That should be fixed to have 1.
1. `getRequestParameter` at this we don't think it should have a third parameter, as the `$blRaw` might be deprecated in upcomming major release. So we can't accept that part of the pull request
1. (optional) provide additional method:  
`public function getRequestRawParameter($sName, $sDefault = null)`  
This method should return raw value of the request.

MantasVaitkunas:
@revoltek-daniel, did you get the last message from @Buratinas? Do you want to change this so we can merge the pull request? Any other opinion? Cheers!

revoltek-daniel:
@MantasVaitkunas i will try to improve this pull request on weekend

ooxi:
Shouldn't we just wait for the [??](https://wiki.php.net/rfc/isset_ternary) operator? ;)

MantasVaitkunas:
@ooxi, I'm afraid that we gonna need to wait for this little bit too long :older_man:

MantasVaitkunas:
@revoltek-daniel, I see you improved the code, how about unit tests? :)

PR #211, opened 2015-02-03
Update genexport_do.php, fix mixed line endings on export

SvenBrunk:
What sense does it make to first replace all line break chars with newline and then output everything with additional carriage return + newline? we should either simply stick to the newline we chose before or make this configurable.


artsev-oxid:
Hi,
Please sign the OCA using the service.

SvenBrunk:
Now did. Thought OXID employees that already hat OCA status in the past would have been transferred to the new system automatically. Sorry.

PR #217, opened 2015-02-28
0006059: strong-tag not closed in listitem_grid.tpl when in b2b mode

adriankirchner:
For further information see https://bugs.oxid-esales.com/view.php?id=6059


ooxi:
:+1:

PR #218, opened 2015-03-01
0006060: user list is not up-to-date when deleting a user from list

adriankirchner:
For further information see https://bugs.oxid-esales.com/view.php?id=6060

### DIFF
--- a/source/application/controllers/admin/user_list.php
+++ b/source/application/controllers/admin/user_list.php
@@ -57,14 +57,12 @@ class User_List extends oxAdminList
     protected $_sThisTemplate = 'user_list.tpl';
 
     /**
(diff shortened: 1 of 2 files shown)


ooxi:
Shouldn't `deleteEntry` just update the list? Changing the behaviour of initializing the list from `init` to `render` should be considered a bc break.

adriankirchner:
Imho there is no need to add extra code to compensate a behavior which is simply not necessary.
Manipulating the item list in `init()` is not necessary and let `deleteEnty()` doing the job with unsetting the deleted user from the already built item list also breaks the OXID standard way of building and displaying lists.
Even when you consider this PR a bc break it shouldn't be a problem since I committed this to `b-dev-ce`.

PR #220, opened 2015-03-02
Cache `getOffset` invokation

ooxi:
`getOffset` returns both `top` and `left` coordinates, thus invoking it twize is unnecessary


kermie:
@ooxi sorry, we didn't see at a glance that the OCA wasn't signed yet (checks presently take up to five or more days). Could you please sign it for the next commits?
https://github.com/OXID-eSales/oxideshop_ce/blob/b-dev-ce/CONTRIBUTING.md

PR #222, opened 2015-03-09
fix bug #6076 -  function oxViewConfig::isModuleActive returns false, if...

a4p-ASD:
... module has no vendor folder

function isModuleActive also returns false, if module does not extend any oxid class; therefore check uses wrong array

now uses 'aModuleVersions' instead of 'aModules' from oxconfig

### DIFF
--- a/source/core/oxviewconfig.php
+++ b/source/core/oxviewconfig.php
@@ -1380,10 +1380,11 @@ class oxViewConfig extends oxSuperCfg
     {
         $blModuleIsActive = false;
 
-        $aModules = $this->getConfig()->getConfigParam('aModules');
+        // use aModuleVersions instead of aModules, because aModules gives only modules which extend oxid classes
+        $aModuleVersions = $this->getConfig()->getConfigParam('aModuleVersions');
 
-        if (is_array($aModules)) {
-            $blModuleIsActive = $this->_moduleExists($sModuleId, $aModules);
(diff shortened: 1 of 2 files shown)


a4p-ASD:
finally fixed the unit test! :-)

MantasVaitkunas:
Hi @a4p-ASD,

thanks for PR, it fixes the issue :+1: Only one small note, could you reopen pull request and make one commit instead of 5 :) I would like that git history would look little bit nicer.

jkrug:
'cmon, this is really not a cool reply. It's so anoying to do PRs for OXID!

MantasVaitkunas:
@jkrug, if there are any problems to reopen PR and fix commit history, we gonna accept this one. As I mentioned it's small thing and bug fix is really good. I'm just saying what's in my opinion could be improved as we all seek better quality, even in git commit history.

jkrug:
@MantasVaitkunas see the other side: All people that are doing PRs are normally doing this beside their work in their free time to help making OXID a better software. And you always find something that stops you from merging.

This is a bug that you don't need to fix anymore. And you get the requested Unittest in 1 day. After that you need 3 weeks to check it and come up with "the git history is not so nice".

There are bugs open for years - seems like we found a reason for that.

a4p-ASD:
Sorry - i'm on vacation... ;-) 

---- Mantas Vaitkūnas schrieb ----

PR #224, opened 2015-03-11
Changed OXID generation for template blocks to prevent multiple entries.

kyoya-de:
A little bugfix to prevent duplicates in OXID's table "oxtplblocks".

The OXID isn't randomly generated anymore. It's now based on the extended template, the extended block, the shop ID and the module ID.

A little limitation could be the fact, that a block in a template could only be extended once in each module.

### DIFF
--- a/source/core/oxmoduleinstaller.php
+++ b/source/core/oxmoduleinstaller.php
@@ -380,15 +380,34 @@ class oxModuleInstaller extends oxSuperCfg
         if (is_array($aModuleBlocks)) {
 
             foreach ($aModuleBlocks as $aValue) {
-                $sOxId = oxUtilsObject::getInstance()->generateUId();
+                $sOxId = MD5($aValue['template'] . $aValue['block'] . $sShopId . $sModuleId);
 
-                $sTemplate = $aValue["template"];
-                $iPosition = $aValue["position"] ? $aValue["position"] : 1;
-                $sBlock = $aValue["block"];
-                $sFile = $aValue["file"];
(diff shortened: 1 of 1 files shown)


MantasVaitkunas:
Hi @kyoya-de,
thanks for pull request. Could you please provide a case when this bug appears? Also would be great if you would register it in bugtrack: https://bugs.oxid-esales.com/main_page.php

kyoya-de:
This appears every time a module, which contains "template blocks" in the metadata.php file, deactivated an re-activated. Then you'll get duplicates in the "oxtplblocks" table.

MantasVaitkunas:
Unable to reproduce. I was using this module: https://github.com/OXID-eSales/paypal and eShop development version. Could you provide me more specific information how I could reproduce the issue?

kyoya-de:
The bug occurs if a module gets partially disabled, e.g. a new extend was added, and it's activated again in OXID admin.

kyoya-de:
There is already a bug-ticket. this pull request can be a solution of the resulting problem.
https://bugs.oxid-esales.com/view.php?id=5925

MantasVaitkunas:
The bug 0005925 is not related. I remember that the issue you are talking about was in earlier eShop versions, but in development version I can't reproduce it... Could you define steps what to do, to reproduce it? It would be best if you would register it in bugtrack system.

MantasVaitkunas:
@kyoya-de, still waiting for reply here. Can we close this PR unmerged?

PR #228, opened 2015-03-25
create export folder if not exists

revoltek-daniel:
create export folder if not exists like mentioned in https://bugs.oxid-esales.com/view.php?id=6094

### DIFF
--- a/source/application/controllers/admin/voucherserie_export.php
+++ b/source/application/controllers/admin/voucherserie_export.php
@@ -113,7 +113,12 @@ class VoucherSerie_Export extends VoucherSerie_Main
      */
     protected function _getExportFilePath()
     {
-        return $this->getConfig()->getConfigParam('sShopDir') . "/export/" . $this->_getExportFileName();
+        $sHttpDir = $this->getConfig()->getConfigParam( 'sShopDir' ) . "/export/";
+        if (!file_exists($sHttpDir) || !is_dir($sHttpDir)) {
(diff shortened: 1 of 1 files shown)


source/application/controllers/admin/voucherserie_export.php
```diff
@@ -113,7 +113,12 @@ protected function _getExportFileName()
      */
     protected function _getExportFilePath()
     {
-        return $this->getConfig()->getConfigParam('sShopDir') . "/export/" . $this->_getExportFileName();
+        $sExportDir = $this->getConfig()->getConfigParam( 'sShopDir' ) . "/export/";
+        if (!file_exists($sExportDir) || !is_dir($sExportDir)) {
```
MantasVaitkunas: What is the reason to use ``!file_exists($sExportDir)`` here. Maybe it would be enough ``!is_dir($sExportDir)``

MantasVaitkunas:
Hi, thanks for PR :+1: Few notes:
I am not sure if getter is proper place to create directory. I would suggest to move this functionality into method ``run``. By the way, unit tests would be appreciated :)

MantasVaitkunas:
@revoltek-daniel, did you see my last comment? I definitely can't merge this pull request with your provided solution. Instead, I proposed to think about expanding the run method. What do you say?

MantasVaitkunas:
@revoltek-daniel, sorry, I can't merge this pull request with provided solution...

PR #230, opened 2015-03-27
fill netto and brutto price in amountPriceList in oxarticle

robinlehrmann:
We need netto and brutto price in our theme for the amount Price List on the details page. (http://www.vkf-renzel.de/Namensschild-Querformat.html)

I thnik it would be great if you fill brutto and netto every time and use isPriceViewModeNetto in template to check if nettoMode is active.

### DIFF
--- a/source/application/models/oxarticle.php
+++ b/source/application/models/oxarticle.php
@@ -3728,7 +3728,7 @@ class oxArticle extends oxI18n implements oxIArticle, oxIUrl
         // trying to find lowest price value
         foreach ($aAmPriceList as $sId => $oItem) {
 
-            $oItemPrice = $this->_getPriceObject();
+            $oItemPrice = $this->_getPriceObject(true);
             if ($oItem->oxprice2article__oxaddabs->value) {
 
                 $dBasePrice = $oItem->oxprice2article__oxaddabs->value;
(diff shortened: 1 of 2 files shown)


source/application/models/oxarticle.php
```diff
@@ -3728,7 +3728,7 @@ protected function _fillAmountPriceList($aAmPriceList)
         // trying to find lowest price value
         foreach ($aAmPriceList as $sId => $oItem) {
 
-            $oItemPrice = $this->_getPriceObject();
+            $oItemPrice = $this->_getPriceObject(true);
```
MantasVaitkunas: Why did you force netto mode?

robinlehrmann:
If the Shop is in netto Mode I get the netto price but I want to get the brutto price and calcluate the netto price manuelly below. So I can assign the netto and the brutto price to the template. In your actually solution you assign the price to the brutto price variable in template if netto Mode is active and not active. So you can use only netto or brutto price in template. We need both (link in my description) 

Maybe I can add a phpdoc description to explain it  in the Code.

MantasVaitkunas:
Well the problem is when I use brutto mode (config blShowNetPrice = false), I get bad results when:
```php
$oItemPrice = $this->_getPriceObject(true);
```
For example if product bruto price is 1€ in frontend I get result 1.19€ because calculation mode is forced to be netto.

Seems like if netto mode would not be forced, everything would work as you wanted.

robinlehrmann:
Okay. Sorry, I missed that. Could you chance it please and merge? I think a unit test is unnessesary.

MantasVaitkunas:
Fixed and merged your PR: https://github.com/OXID-eSales/oxideshop_ce/commit/626342eaa48b0f72133db43b9606b2c4f471e2a7
Also added unit test: https://github.com/OXID-eSales/oxideshop_ce/commit/7187bc3cf51d5c4db3982b44bb51c24574559484

PR #231, opened 2015-04-15
Fixed access to target Category when loading products

dmersiowsky:
Though a target category is given as function parameter, the "active category id" was loaded directly from Request in line 379. This did not make sense at all. The system does not behave as expected here, if the list controller is not used in standard (OXID-)case.

After some short tests, this change should not cause any problems.

### DIFF
--- a/source/application/controllers/alist.php
+++ b/source/application/controllers/alist.php
@@ -376,8 +376,8 @@ class aList extends oxUBase
         } else {
             $aSessionFilter = oxRegistry::getSession()->getVariable('session_attrfilter');
 
-            $sActCat = oxRegistry::getConfig()->getRequestParameter('cnid');
-            $this->_iAllArtCnt = $oArtList->loadCategoryArticles($sActCat, $aSessionFilter);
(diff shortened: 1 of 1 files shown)


stasiukaitis-saulius:
Hello Daniel,

Thank you for this pull request. Indeed it looks like it could solve some possible issues so we decided to take it. 
It would be very nice from you if you could provide some business case about this fix for us to better understand it.

dmersiowsky:
Thanks for considering a pull. In my special case, I wanted to display a specific category outside the categories view. So I created an aList object and set the desired category as active category (and assign that object to a smarty template for displaying the list). 

Without the fix discussed here, the previously set category would not be considered, but aList would try to reload it from the request. Here's a corresponding code snippet to reproduce the problem:

```php
    /** @var \oxCategory $cat */
    $cat = oxNew('oxCategory');
    $cat->load($category_id);

    /** @var \aList $view */
    $view = oxNew('aList');
    $view->init();
    $view->setActiveCategory($category);
    $view->setListType('infogrid');
```

PR #232, opened 2015-04-16
Fix Bug #4868 false VAT and shipping cost after recalculation- set Address Id in session variable

osaprog:
Fix Bug #4868 false VAT and shipping cost after recalculation

### DIFF
--- a/source/application/models/oxorder.php
+++ b/source/application/models/oxorder.php
@@ -1429,6 +1429,9 @@ class oxOrder extends oxBase
         // setting stock check mode
         $this->_oOrderBasket->setStockCheckMode($blStockCheck);
 
+        // set selected address in Session Variable
+        $this->getOrderUser()->setSelectedAddress();
+
         // setting virtual basket user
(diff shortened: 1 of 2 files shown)


stasiukaitis-saulius:
Hello  @osaprog,

I tried your pull request locally but it did not work for me as expected.

What I did:
I created a new shop with demo data as if it would be in Germany.
I enable addresses from Germany and USA.
Germany is local country so VAT should be applied to order.
USA is foreign country so VAT should not be applied to order.
I create an order as German customer.
Order was created with German VAT.
I logged in to admin and change user billing address to USA in order address tab.
Order VAT did not change. I expected that order VAT would change to zero.
I save order information in main tab.
Order VAT changed to 0 as expected. Same behavior existed before your pull request.

Probably I miss something. Can you add what work-flow you use when this bug fix works?

MantasVaitkunas:
@osaprog, I can't merge the provided solution: please see @stasiukaitis-saulius comments + Travis and CLA fails.

PR #240, opened 2015-06-16
Added support for PSR0 and PSR4 autoloading.

keywan-ghadami-oxid:
This brings support for namespaces and makes it possible to include other 3rd party libaries to core.

This pull request was designed to be backward compatible, so i tried not to change old code as much as possible
to lower the risc of breaking bc breaks or new bugs or unmasking bugs.

PSR3 Logger Interface and a Logger Implementation will be part of an other pull request,
also not included in this merge request is PSRR3/PSR4 autoloading support for modules.

Performance:
- negative impact should be ver low (hopyfully not mesureable),
- positve impact depends on usage of namespaces.

I am also thinking of caching $aClassPaths in the autoloader to make performance improvements but this would also be an other merge request.

PR #241, opened 2015-06-16
Feature/psr3 logger interface

keywan-ghadami-oxid:
this adds the PSR3 (Logger) support with all interfaces with all classes and traints
all classes that are instanciate via oxnew can get the logger object injected
they only need to implement Psr\Log\LoggerAwareInterface.

e.g.:

```php
class Start extends oxUBase implements \Psr\Log\LoggerAwareInterface
{
    use \Psr\Log\LoggerAwareTrait;
//...
    public function render()
    {
        $this->logger->alert("oxid rocks cause oxid logs");
```

this pull request depends/includes https://github.com/OXID-eSales/oxideshop_ce/pull/240


MantasVaitkunas:
This PR is dependent on #240 (that just was refused). Sorry, can't merge it.

PR #242, opened 2015-06-24
Bugfix payment discount only for discountable products

web-architektin:
negative payments costs are also calculated for non-discountable products and categories. We think, this is not correct.

### DIFF
--- a/source/composer.lock
+++ b/source/composer.lock
@@ -652,12 +652,12 @@
             "source": {
                 "type": "git",
                 "url": "https://github.com/OXID-eSales/testing_library.git",
-                "reference": "865416266619f5ab8774f3d5c3b9cc2f2d4e64cd"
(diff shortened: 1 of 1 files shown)


MantasVaitkunas:
Thanks for your PR. I have to take a closer look at it. Hang on. 
In the mean time, could you please let me know the bug number, did you find it in the bug tracker? Cheers.

MantasVaitkunas:
Thanks for your pull request, @web-architektin, but we can't merge it.
What you try to achieve is a "misuse" of this calculation. A discount should always be set in discounts but - although it is possible, a negative surcharge for the payment can hardly be fully handled as a discount, can't it?
Just a hint, if you need it in one of your projects: strictly try to work with unit tests. Especially in this case I see a high potential of appearing of other unwanted behavior if you fix it the way you provided it.

PR #243, opened 2015-07-13
Delete .phpstorm.meta.php because of native PhpStorm plugin

Haehnchen:
I released a native PhpStorm plugin which handles this in a smarter way. see https://plugins.jetbrains.com/plugin/7853. Issues and feedback welcome: https://github.com/Haehnchen/idea-php-oxid-plugin


ooxi:
Could you post an example of a `.phpstorm.meta.php` file generated by your plugin?

Haehnchen:
this is a native plugin, so it simply not exists :). see https://github.com/Haehnchen/idea-php-oxid-plugin/blob/master/src/de/espend/idea/oxid/types/OxidFactoryTypeProvider.java for implementation.
it also combines the factory pattern with plugin metadata classes eg `extends` .

ooxi:
oic, great work!

adriankirchner:
@Haehnchen really good job! I'm using your plugin since you created this PR and so far it is doing great. Thanks a lot!

I guess you need to sign the CLA via clahub in order to make this PR mergeable.

Maybe you or @kermie could do a post on the dev mailinglist regarding this topic ;) I think there are many oxid devs with PhpStorm out there.

kermie:
@adriankirchner right, will write a note about this great job as a blog post. And yes, CLA would be appreciated.

PR #244, opened 2015-07-23
avoid undefined index notice

revoltek-daniel:
avoid undefined index notice in getModulePath


MantasVaitkunas:
Thanks for PR. We will check this soon. Hang on!

PR #251, opened 2015-08-21
remove hardcoded template name

revoltek-daniel:
remove hardcoded template name to make change of template name easier.


MantasVaitkunas:
@ooxi, your note is absolutely correct. Would be great for upcoming pull requests; but for now I think we are going to accept #251 as it is to keep things as easy as possible.

PR #253, opened 2015-08-27
add some typehints

revoltek-daniel:
add tyehints to get better IDE completion and get rid of strict warnings if method is extended


MantasVaitkunas:
@revoltek-daniel, sorry, but we will have to decline your pull request because of reasons described in this blog post: http://radify.io/blog/type-hinting-in-php-good-or-bad-practice/
If you want to resolve IDE auto completion issue, I would suggest to make another pull request and define types in phpDoc comments only, like you did it now:

```php
/**
* ...
* @param oxUser $oUser
* ...
*/
```

PR #254, opened 2015-08-31
Fixed manufacturer ID comparison.

kyoya-de:
This pull request solves the bug ticket 6221.
http://bugs.oxid-esales.com/view.php?id=6221

### DIFF
--- a/source/application/models/oxmanufacturerlist.php
+++ b/source/application/models/oxmanufacturerlist.php
@@ -123,7 +123,7 @@ class oxManufacturerList extends oxList
         foreach ($this as $sVndId => $oManufacturer) {
 
             // storing active manufacturer object
-            if ($sVndId == $sActCat) {
+            if ($sVndId === $sActCat) {
                 $this->setClickManufacturer($oManufacturer);
             }
 


MantasVaitkunas:
@kyoya-de, thanks for pull request. Just tried to reproduce bug in same way as it was described in bugtrack and add your fix.
In case when manufacturer ID is 10: ``$sVndId`` becomes integer which is equal to ``10``, so ``10 !== "10"``
This means, that when I select manufacturer which has an id 10, I will not see any product in frontend.
So this PR should be improved. Maybe you should convert ``$sVndId`` to string. 

P.S. I would love to see unit tests for this fix.

jkrug:
It's just suggestion from us. Feel free to improve.

floeschie:
I think it is a good idea to avoid implicit type conversions (see also [OWASP PHP Security Cheat Sheet](https://www.owasp.org/index.php/PHP_Security_Cheat_Sheet#Weak_typing)).

However in this particular case developers might end up scratching their heads in case they do not use the default OXID's, but some custom ID's, e. g. from an external system like ERP.

I will talk to the team about this in the nest Sprint Planning. In this case type casting might be a good idea, and I would also suggest to project developers to build quality in to their pull requests – it might at some point save their precious project time.

Please stay tuned for the discussion results.

Sieg:
Hello @kyoya-de, Thank you for your pull request!. I merged it a while ago.

PR #255, opened 2015-09-09
oxnew parameters are serialized and so they can hit servers memorylimit

keywan-ghadami-oxid:
this commit fixes a problem that happens if module-developers use oxNew with heavy constructor arguments (e.g. objects or arrays).
my modification checks the arguments first, only simple values (scalars/null) are accepted for cachable objects.

This commit does NOT fix the problem if someone uses a very big string as constructor argument,
but therefor strlen of the args could be checked additional.

### DIFF
--- a/source/core/oxutilsobject.php
+++ b/source/core/oxutilsobject.php
@@ -157,6 +157,15 @@ class oxUtilsObject
         $blCacheObj = $iArgCnt < 2;
         $sClassName = strtolower($sClassName);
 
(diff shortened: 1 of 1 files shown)


Sieg:
Hey Keywan,
thanks for this pull request. I had a glance on your code, and I think the idea is pretty good. Need to analyse it a bit deeper, hang on.

MantasVaitkunas:
Need to run performance tests on it.

PR #257, opened 2015-09-24
Update oxseodecoder.php

TigerLeon:
In some cases you want to know, if the same URL exists in other shops. For example if you don't have a responsive shop and you want to have a link on your mobileshop which links to the same page in your desktopshop. This extension of the method could help with this case.


Sieg:
Thanks for your pull request, @TigerLeon. Unfortunately, we can’t accept features for the Community Edition that already exist in Enterprise Edition (multi-client capability, subshops). Please also see:

http://wiki.oxidforge.org/OXID_Contribution_and_Contributor_Agreement_FAQ#What_should_I_do_before_I_contribute_code_to_OXID_eShop.3F

PR #259, opened 2015-09-30
add feature: hidden articles

proudcommerce:
for website content / seo its very important to keep seo-urls, also from articles which are not longer available. with this feature its possible to hide articles in article lists, but access articles details page.


proudcommerce:
any status here?

source/application/models/oxarticle.php
```diff
@@ -555,6 +555,9 @@ public function getActiveCheckQuery($blForceCoreTable = null)
         // check if article is still active
         $sQ = " $sTable.oxactive = 1 ";
 
+        // check if article is hidden
+        $sQ .= " and $sTable.oxhidden = 0 ";
```
Crease29: Maybe you schould change this to an OR clause for better usability.
I just see the case where someone wants to make an article visible to a collegue and thinks: "Okay it shouldn't be active for everyone in but hidden."
Maybe you should enclose this line with a check if the current user is an admin?

floeschie:
I like this feature, however @Sieg can you check the performance impact before adding this to core?

source/application/models/oxarticle.php
proudcommerce: i think this is an special feature and not the normal usecase

proudcommerce:
open since 5 weeks :-(

MantasVaitkunas:
Hi,
yep. As @floeschie wrote, we would like to run performance tests first. Due to an issue with internal testing environment we can't presently. Will get back to this one as soon as possible.

source/application/models/oxarticle.php
```diff
@@ -555,6 +555,9 @@ public function getActiveCheckQuery($blForceCoreTable = null)
         // check if article is still active
         $sQ = " $sTable.oxactive = 1 ";
 
+        // check if article is hidden
```
keywan-ghadami: Iwould like to have an cosmetic change here, i think it's better document the "why" instead the "what" e.g.
//  hidden articles should not be included in lists e.g the category pages
Btw +1 for this pull request

Shiftas:
Hi, just checked performance impact and there seems to be none. Could you rebase this branch so we could accept this pull request?

PR #261, opened 2015-10-03
RDF Meta tags evaluation

jsmunich:
gr:description should first be evaluated, then tags should be stripped. 

otherwise, if the description contains `getCurrentHomeDir()}]/some-file.png" />`, it fails to produce valid html code, as `->` seems to be counted as a closing tag.


oxidsupport:
Is this related to the RDFa part of [this issue](https://bugs.oxid-esales.com/view.php?id=6234) in the bugtracker?

Sieg:
Hello @jsmunich, Thank you for the fix, i just merged your pr.

PR #262, opened 2015-10-08
Update oxarticle.php

TigerLeon:
The performance-setting bl_perfLoadAttributes has no functionality at the moment. This change should give it a sence.

### DIFF
--- a/source/application/models/oxarticle.php
+++ b/source/application/models/oxarticle.php
@@ -2473,8 +2473,11 @@ class oxArticle extends oxI18n implements oxIArticle, oxIUrl
     public function getAttributes()
     {
         if ($this->_oAttributeList === null) {
-            $this->_oAttributeList = oxNew('oxattributelist');
-            $this->_oAttributeList->loadAttributes($this->getId(), $this->getParentId());
+            $this->_oAttributeList = false;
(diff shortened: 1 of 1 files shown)


kermie:
OCA is signed, although not checked here automatically.

leofonic:
getAttributes isn't executed automatically, so switching it off has no benefit. Why not remove this setting?

keywan-ghadami-oxid:
i am wondering if the default value for bl_perfLoadAttributes is false and so this change would break the functionality in existing shops.

MantasVaitkunas:
Hi @TigerLeon,
have a look at the tests failing: https://travis-ci.org/OXID-eSales/oxideshop_ce/jobs/84253629#L726. Apparently, your suggestion doesn't pull out any variants in case they're needed. To be honest: I would rather go for @leofonic 's proposal - remove what is not needed any more (as there were getters introduced ages ago for this behavior.

Shiftas:
Removed this Performance option from shop. https://github.com/OXID-eSales/oxideshop_ce/commit/00e9b3438c1821ad5583f94a5ef8abe80d8c5717

TigerLeon:
Thank you. Also needed in EE:
https://bugs.oxid-esales.com/view.php?id=6239

PR #263, opened 2015-10-08
Cleanup JSDoc and JavaScript warnings

marvinhagemeister:
+ Add JSDoc where it was missing.
+ Remove unnecessary global variables because of missing `var` keyword.
+ Use `[]`-assignment for arrays instead of `new Array()`.
+ Cache obvious jQuery object usage.
+ Simplify some return statements (those that PHPStorm automatically gets right).
+ Fixed indention of some files


ooxi:
This is a huge commit consisting of multiple logically unrelated issues. It would be preferable to split them into logical units (much like your list of things fixed)

marvinhagemeister:
Thanks for the feedback @ooxi . You're right,  I'll split up the PR into smaller ones.

ooxi:
Otherwise :+1: :)

PR #265, opened 2015-10-09
Use array literals for assignments instead of 'new Array()' in JavaSc…

marvinhagemeister:
Using `var foo = [];` is more common than `var foo = new Array();` and has the benefit of being shorter. There is no difference in behavior between the two.


ooxi:
:+1:

Sieg:
@marvinhagemeister, Thank you for your pr. Merged :)

PR #267, opened 2015-10-10
Use CSS shorthand properties where possible

marvinhagemeister:
Is a bit nicer to read and avoids overwriting of previously declared attributes (as is the case with `padding` and `margin`).

### DIFF
--- a/source/out/azure/src/css/oxid.css
+++ b/source/out/azure/src/css/oxid.css
@@ -81,8 +81,7 @@ ul ul, ol ul {
 }
 
 ol li {
-    list-style-type: decimal;
-    list-style-position: inside;
+    list-style: decimal inside;
 }
 
 ol ol {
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
I'm honestly baffled why travis fails, because the failing test seems to be totally unrelated :s

ooxi:
I agree, the Travis build error seams to be network releated (``ConnectionException`). Nevertheless I advice against using CSS shorthand properties because of the reduced reability. For example:

**before**

    margin: 0;
    margin-left:10px;

**after**

    margin: 0 0 0 10px;

Even without knowing CSS to the heart I can read and interpret the first statement without any problems whereas the second one needs additional knowledge or tooling support. Compression can better be done by automatic tools and should be left to them. The source should remain in the preferable form for modification.

:+1: for the stylefixes though, newlines and intendation :)

marvinhagemeister:
I see your point, but I don't think that this it is a good practice to declare properties twice nonetheless. Any decent css-linter will throw a warning here. By improving the css code structure itslef one could easily avoid having to reset all the other values to 0, but this is for later.

For now I'll only fix the warnings and errors, so that we have an error free base to improve on without having to dig through dozens of warnings first.

Sieg:
@marvinhagemeister, Thank you! Merged.

PR #268, opened 2015-10-12
Feature/pr persparams handling

a4p-ASD:
please see mantis #6244


Sieg:
Hello, we like your idea about collecting all the "persParams" related blocks to one place, but to approve the pull request, could be great if you will improve the quality of the implementation:
       
* Remove all moved code from original templates.
* In place of introduced input.tpl and output.tpl files we want this blocks to be just in separate directory as separate files for every extracted block, so respectively you can remove the conditional parameter in your includes.

a4p-ASD:
Hello,
sure, i would clean up the moved code and comments ...

But, if i understood your last request right, why do you want the extraced blocks in seperate files?
That was my main goal for this PR, to have all the persparams at one place/in one block (ok, two).
For example:
i have a module, which adds a new persparam for an article
instead of displaying only the value of the persparam, i need to add custom html and smarty stuff (or use a widget)
-> i would again need to change every file/block to display this custom persparam...

With a single file for the output of the persparam: i could do it like this:
-> extend block "persparam_output" with module
block content:
    [{ $smarty.block.parent }]
    [{if $sPersParamKey == "myCustomPersparam" }]
        [{* the output is the same for every tpl /block ... *}]
        <b>my persparam</b>: [{include file="myModule/persparam...tpl }]
        ...

Could you please explain more what you meant/want with "we want this blocks to be just in separate directory as separate files"?

Sieg:
Hey! Its about lowering the complexity of the structure. By separating blocks (every little case in condition) to files you can remove the conditions introduced in input and output files, and remove the parameter used for these conditions in include statements. If you will do so, the "include" itself will be the condition which already takes the block you need.

So from the example:
```javascript
[{block name="persparams_input"}]
    [{if $oView->showPersParam($sPersParamKey)}]
        [{if $tpl == "page_details_inc_productmain"}]
            [{if $tpl == "page_checkout_inc_basketcontents_editable"}]
                <p>...</p>
            [{/if}]
        [{else}]
            [{if $tpl == "page_checkout_inc_basketcontents_editable"}]
                <input type="hidden" name="aproducts[[{$basketindex}]][persparam][[{ $sPersParamKey }]]" value="[{ $sPersParamValue }]">
            [{/if}]
        [{/if}]
    [{/if}]
[{/block}]
```
My suggestion will be to extract a new file with name pers_params/page_checkout_basketcontents.tpl :
```javascript
[{block name="page_checkout_basketcontents"}]
    [{if $oView->showPersParam($sPersParamKey)}]
        <p>...</p>
    [{else}]
        <input type="hidden" name="aproducts[[{$basketindex}]][persparam][[{$sPersParamKey}]]" value="[{$sPersParamValue}]">
    [{/if}]
[{/block}]
```
So you can write include without additional parameters and easily overwrite the little block you need, but not the whole huge input.tpl block. Everything will still be in one place - separate directory.

a4p-ASD:
OK, i accept that as a good compromise!

Your right, my persparam_input tpl was a bit too ugly...

For some modules, it can be enought to extend oView and change the new persparam functions - instead of the blocks...

Please give me some time to change the PR - i hope its done before christmas vacation ;-)

How do i do that at best with github - shall i delete the old PR?

stasiukaitis-saulius:
Hi @a4p-ASD,

You can simply add one more commit to your branch and it will automatically ends up in this pull request. You can even rebase and change your previous commit to have nicer history. In general pull request is just a link to your branch.
For more information read this: http://stackoverflow.com/questions/10147445/github-adding-commits-to-existing-pull-request

Cheers!

a4p-ASD:
Hi,
i´ve added a new commit. Hope its better and cleaner now... ;-)
What do you think?
Greetings!

a4p-ASD:
The PR is working on CE 4.9.5 ...  all looks good now.

source/application/views/azure/tpl/page/account/order.tpl
```diff
@@ -67,9 +67,14 @@
                                         [{ $orderitem->oxorderarticles__oxtitle->value }] [{ $orderitem->oxorderarticles__oxselvariant->value }] <span class="amount"> - [{ $orderitem->oxorderarticles__oxamount->value }] [{oxmultilang ident="QNT"}]</span>
                                     [{if $oArticle->oxarticles__oxid->value && $oArticle->isVisible() }]</a>[{/if}]
                                     [{foreach key=sVar from=$orderitem->getPersParams() item=aParam}]
+[{*---------------------------------------------------------------------------------------------*}]
+[{*
```
alfredbez: You need to clean up these lines, too.

source/core/oxview.php
```diff
@@ -1017,4 +1017,60 @@ public function getSidForWidget()
...
+     * @param $s_paramKey
+     * @param $s_paramValue
+     *
+     * @return null|other
+     */
+    public function getPersParamValue($s_paramKey, $s_paramValue) {
+
+        $_ret = null;
+
+        if ($s_paramKey == "details")
```
alfredbez: How can you get a Persparam-Value with an other key then 'default'? (see also 6f3675413d67305d285e2173238a220bccc8b734)

source/application/views/azure/tpl/page/account/order.tpl
a4p-ASD: forgotten... Thank you!

source/core/oxview.php
a4p-ASD: Your right.
My thought was, the function is intended to be overloaded by modules... and the module returns "custom" values or no value or whatever ...
But at "default", it should return every persparam value.

PR #269, opened 2015-10-12
Fix 'oxid' checks in admin controllers renders.

vaidasif:


ooxi:
Could you provide more context? What problem does your patch fix?

vaidasif:
Mainly at admin controllers are used '-1' value to identify new object. There was just '-', so I thought that it should be same in all admin controllers.

Sieg:
Hi @vaidasif, Thanks for pull request. Merged.

PR #270, opened 2015-10-12
Check if object is set before comparison.

vaidasif:


Sieg:
Hello @vaidasif, Thank you for the pull request. Merged.

PR #271, opened 2015-10-12
Use ceil() instead of round() using hardcoded values for pagination.

vaidasif:


ooxi:
:+1: `ceil` is much easier to understand

PR #272, opened 2015-10-12
If no data for search were passed - return as fast as possible.

vaidasif:


Sieg:
Hello @vaidasif, Thanks for pull request. Merged.

PR #275, opened 2015-10-13
User exceptions translations

vaidasif:


Shiftas:
Hi, could you use different keys for testing admin translations in testTranslateStringIsAdmin() instead of removing test case? Also EXCEPTION_USER_USERCREATIONFAILED and 
EXCEPTION_USER_NOVALUES usages should be changed to ERROR_MESSAGE_USER_USERCREATIONFAILED and ERROR_MESSAGE_USER_NOVALUES instead of adding new translation keys.

PR #280, opened 2015-10-23
Simplify and refractor oxEqualizer

marvinhagemeister:
Simplify and refractor oxEqualizer.

+ Remove unused jQuery-UI widget factory. `oxEqualizer` is only used directly
+ Use `Array.map()` instead of a custom loop (polyfill for <IE9 included)
+ Remove unused oxEqualized CSS-Class
+ Declare `getTallest()` before it is used

Note: This plugin will be redundant when support for IE9 is dropped (may take some years still), because Flexbox solves all of the typical problems that occur with floats in CSS.


ooxi:
I'm really opposed to modifying core JavaScript objects, wouldn't it be sufficient to check if `Array.map` exists and if not use a namespaced polyfill?

marvinhagemeister:
The way you described it is exactly how my PR works. The file IE9.js is only loaded for old versions of IE which don't support Array.map() anyway. The check happens via a conditional comment and I think there is no need to duplicate it inside js. All other browsers have support since the beginning (the only exception beeing firefox which added support in 1.5 and that is way beyond ancient).

ooxi:
I understand that the polyfill will affect IE 9 but there it might break user code which wants to use another `Array.map` implementation

adriankirchner:
@ooxi they can use their own implementation if the prototype it before the IE9.js: https://github.com/marvinhagemeister/oxideshop_ce/commit/2ec9b096b4ab8aa22e5ff617739d8f9d8ad414c0#diff-0cba9049c5efe401104fa7fd364c241fR10

@marvinhagemeister :+1:

marvinhagemeister:
Well if someone willingly overwrites the implementation defined in the spec than he'll better know what he is doing. I don't think we need to support such a use cases where someone is non-complient with the official spec.

Shiftas:
Hi, I think it is great what you did, but could you also move Array.prototype.map definition to out/src/js/IE6-8.js and rename current IE9.js file to selectivizr.js as it only contains this library and is placed in libraries folder?

marvinhagemeister:
@Shiftas Ah yes, that's a great suggestion and cleaner than the current approach. I'll update the PR.

PR #290, opened 2015-11-05
fix broken javascript in combination with varnish

keywan-ghadami-oxid:
oxscript must pack some javascript into a string to execute it in a global scope, for that

some characters must be replace to generate a valid string.

In the old version it does only replace quotes but newlines are also problematic and must be escaped.

Without that fix you will get javascript errors in combination with varnish in some situations, please contact me if you need to reproduce that.

Implementation details: i used a extra method, which i copied from a async javascript support module that we are using in projects.

### DIFF
--- a/source/core/smarty/plugins/function.oxscript.php
+++ b/source/core/smarty/plugins/function.oxscript.php
@@ -186,7 +186,7 @@ function _oxscript_execute( $aScript, $sWidget, $blAjaxRequest )
     if (count($aScript)) {
         foreach ($aScript as $sScriptToken) {
             if ( $sWidget && !$blAjaxRequest ) {
-                $sScriptTokenSanitized = str_replace( '"', '\"', $sScriptToken );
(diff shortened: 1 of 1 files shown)


source/core/smarty/plugins/function.oxscript.php
```diff
@@ -197,6 +197,12 @@ function _oxscript_execute( $aScript, $sWidget, $blAjaxRequest )
     return $sOutput;
 }
 
+function _oxscript_sanitize($sScripts)
+{
+    $sScriptTokenSanitized = strtr($sScripts,['"' => '\"',"\r" =>'',"\n"=>'\n']);
```
adriankirchner: Since oxid still supports PHP 5.3 [1] you need to use the old array() syntax instead of the short array syntax introduced in PHP 5.4.
Just commit the change to your branch *keywan-ghadami-oxid:fix-oxscript-newline* and it will be automatically applied to this pull request.

[1] http://wiki.oxidforge.org/Downloads/4.9.6_5.2.6

PR #292, opened 2015-11-06
Bug fixing #6134: set setup lang files (EN and DE) to UTF-8

kermie:
Check this one out. Hope it is fine to fix #6134.


MantasVaitkunas:
Pull request does not really fix the problem, as without it problem is only with PHP 5.6 and with pull request PHP 5.6 is fine, but 5.5, 5.4 and 5.3 is not working. We will fix this ASAP.

PR #295, opened 2015-11-21
fixed various PhpDoc issues to improve code inspection/navigation in IDEs

chriz28:


Shiftas:
@chriz28 thanks a lot for your commit! We would like to accept everything but the @var comments as there's a really cool PhpStorm plugin available which works much better than we could achieve with this comments: https://plugins.jetbrains.com/plugin/7853. That's why we decided to not use it any more.
Can you please split your pull request the way we can simply merge it?
Also, we are still missing your signed OCA which is policy.

revoltek-daniel:
@chriz28  if you can't or won't use the plugin, try this gist: https://gist.github.com/revoltek-daniel/748ff6d8543405fb2c2a (its smaller as the generated .phpstorm-oxid.meta.php and supports multiple namespaces) to resolve parent classes and use dynamicReturnType plugin (https://plugins.jetbrains.com/plugin/7251?pr=)

kyoya-de:
The PhpStorm plugin has one big issue: It works only if the shop is installed on the local machine. If the shop runs inside a VM (which is the preferred way to develop), it can't access the DB.
There is also a nice CLI tool which can be exeuted on the machine where the shop runs. Take a look at https://github.com/marcharding/oxrun

Sieg:
Sorry, can't merge this pull request as OCA isn't signed.

chriz28:
Hey,

sorry I didn’t answer for so long. At first I was too busy, then I somehow forgot about it. I saw that you rejected the pull request in the meantime, which is ok since you haven’t heard from me. I might make another one soon, this time without the @vars.

But what is an OCA?

Greetings, 
Christian

stasiukaitis-saulius:
@chriz28 OCA stands for OXID Contributor Agreement. You should see that  Contributor License Agreement is not signed near this pull request.
By the way I see you add comments near oxNew for type hinting. We do not do this anymore as there is plugin which solves this issue https://plugins.jetbrains.com/plugin/7853?pr=

MantasVaitkunas:
PR not signed for weeks. Sorry have to close it again.

PR #301, opened 2015-12-10
skip empty vouchers gerneration

revoltek-daniel:
Its possible in Oxid Admin Backend to generate empty voucher codes in Admin Backend. Choose custom code and set an amout of codes and click generate.

This codes are not useable cause javascript frontend validation requiered a value.

This fix skip generation of empty codes.

### DIFF
--- a/source/application/controllers/admin/voucherserie_main.php
+++ b/source/application/controllers/admin/voucherserie_main.php
@@ -148,6 +148,14 @@ class VoucherSerie_Main extends DynExportBase
      */
     public function start()
     {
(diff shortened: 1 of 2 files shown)


oxidsupport:
related issue in the bugtracker: https://bugs.oxid-esales.com/view.php?id=6284
So the question is where to fix it: In the admin panel, in the basket controller or both?

revoltek-daniel:
@oxidsupport in the Admin panel, cause if you can't generate wrong codes, you can't redeem them.

PR #303, opened 2015-12-22
canonical url for content controller

Alpha-Sys:
At the moment there is no canonical url on content pages because the function getCanonicalUrl is missing in the content controller. So I added the function getCanonicalUrl in the content controller.


Sieg:
Thanks @Alpha-Sys, just merged your pr.

PR #308, opened 2015-12-30
php 7 - fix incorrect jpgraph php version check

adriankirchner:
The jpgraph library uses a custom function to check the php version requirement (>= 5.1.0) which can't handle php version numbers != 5.
Using the version_compare() function lets the corresponding unit test succeed. Graph generation still works in php 7.
This is the last necessary change to let the whole testsuite succeed with php 7 (in half the time and with just 50 percent memory usage compared to 5.x)


ooxi:
:+1:

PR #310, opened 2016-01-02
0006304: oxseohistory is left untouched when deleting entities like oxArticle, oxCategory, oxContent, oxManufacturer, oxVendor

adriankirchner:
For further information see https://bugs.oxid-esales.com/view.php?id=6304

### DIFF
--- a/source/application/models/oxseoencoderarticle.php
+++ b/source/application/models/oxseoencoderarticle.php
@@ -645,6 +645,7 @@ class oxSeoEncoderArticle extends oxSeoEncoder
         $sIdQuoted = $oDb->quote($oArticle->getId());
         $oDb->execute("delete from oxseo where oxobjectid = $sIdQuoted and oxtype = 'oxarticle'");
         $oDb->execute("delete from oxobject2seodata where oxobjectid = $sIdQuoted");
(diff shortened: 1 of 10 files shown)


Shiftas:
Hi, could you rebase this branch so we could accept this pull request?

adriankirchner:
Hi, I need some time to get back to this topic.

keywan-ghadami-oxid:
Hi adrian,
do you remember why it was needed to remove this entries from seo history, is there a problem with this entries? 

I am asking because i think the information that an url was valid in the past is important for some seo related tunings. Some seo agencies like to see status 410 (gone) to be send, some other like see an permanent redirect to an search result page. 

Btw. i see a need to refactor that stuff sooner or later because it is not DRY (by an factor of 5) which already caused Adrian to repeat him self five times in that code. maybe this is a good time to extract things in a parent class (or a delegate).

Shiftas:
@adriankirchner, we want to merge it as it is - @keywan-ghadami's ideas are good but actually belong to another story. Still - can you please re-base so we can quickly merge? Cheers!

MantasVaitkunas:
@adriankirchner, apparently you are busy. Going to merge it manually.

PR #313, opened 2016-01-19
performance can be improved by using intervalls and REQUEST_TIME

keywan-ghadami-oxid:
by replacing the current time with a rounded request time it's possible for the querycache to cache the results of an sql query.

I used a 5 minutes interval for rounding so all queries can be cached for at least 5 minutes,
that mean if an article is configured to become active at 14:02 the oxid shop will starting showing it at  14:05. I guess this should be documented but shouldn't be problematic for normal usecases.

maybe you have arguments why 5 minutes is to much, but at least one minute intervals
can be done without bc break and additional efforts for making it configurable.

```
    //round up 1 minute so query cache can work
    $timestamp = ceil($timestamp/60)*60;
```
I did no performance test on this, would be great to get some feedback if it helps.


keywan-ghadami-oxid:
Unittest are failing because they test for the old sql string, but they must use the same new logic to run successful, do you need that unittest fix from me?

keywan-ghadami-oxid:
creating new pull request against master

PR #314, opened 2016-01-25
maybe shop is a littel bit faster without profiling.

keywan-ghadami-oxid:
no tests where done so far, i hope this is measurable.

To test please set debug level in config.inc.php to 0


Shiftas:
Yes, might have been. We just profiled it and came to the conclusion that profiling takes nearly the same effect for performance as the wished effect could achieve. Decided to not merge it. Thanks for your effort anyway!

PR #315, opened 2016-01-27
Fix � in Setup

alfredbez:
before:
image

after:
image


kermie:
Is it PHP 5.6 related? Am not sure if not the header has to be adapted as well.

ooxi:
Yea without changing of the HTTP Content-Type header or HTML meta charset, the change cannot work in all browsers.

alfredbez:
> Is it PHP 5.6 related?

I've tested this in two different vagrant boxes (5.6 and 7.0) and both show me the same result.

HTTP Content-Type is `Content-Type: text/html; charset=UTF-8` (in Chrome DevTools)

Crease29:
The charset header is UTF-8 because it's the default charset header since PHP 5.6.
What about PHP versions < 5.6?

keywan-ghadami-oxid:
@Crease29 so you suggest to add 
```php
//explicitly set utf8 for php < 5.6
header('Content-Type: text/html; charset=utf-8');
```
into
https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Setup/index.php

Crease29:
Exactly @keywan-ghadami-oxid. Wouldn't it be the best solution to switch OXID eShop completely to UTF-8, nowadays?

keywan-ghadami-oxid:
@Crease29 we already did some steps to go that way, we always have to consider things like performance and compatibility, but i totally agree that we should keep on moving to utf8 everywhere.

Shiftas:
Well, the suggested solution would work properly for use with PHP 5.6 and higher. Unfortunately, it will break for use with PHP 5.5 and lower so we have to find another solution that fits all PHP versions. We'll merge this pull request and add the universal solution for it.

keywan-ghadami-oxid:
for those interested in technical details of the chosen implementation:
- the setup sends know http header with utf8,
- and setup language files are also switched to utf8 encoding,
- html meta tag was used to force utf8.  
-- more details about that tag: http://stackoverflow.com/questions/4696499/meta-charset-utf-8-vs-meta-http-equiv-content-type
see also other commits
https://github.com/OXID-eSales/oxideshop_ce/commit/562e79f2b44d405c5bfaded863fd37ac23763668

PR #317, opened 2016-02-03
this fix method isCurrentUrl (for modules that handle request directly)

keywan-ghadami-oxid:
this fix method isCurrentUrl so it can recognize the current url if the request goes directly to an php file in a module (that may call the oxid bootstrap by it self)

this is a blocker for me, because some modules already do handle requests directly by there self, and the bootstrap will get the wrong shop id in some situations.


keywan-ghadami-oxid:
http://bugs.oxid-esales.com/view.php?id=6346

PR #318, opened 2016-02-03
performance can be improved by using intervalls and REQUEST_TIME

keywan-ghadami-oxid:
by replacing the current time with a rounded request time it's possible for the query cache to hit the an cached sql query.

I used a 1 minutes interval for rounding so all queries can be cached(hit) for at least 1 minutes,

I did no performance test on this, would be great to get some feedback if it helps.
To test the "active from/to" stetting must be enabled, and you need high load situations on specific category pages.

### DIFF
--- a/source/Application/Model/oxarticle.php
+++ b/source/Application/Model/oxarticle.php
@@ -576,8 +576,7 @@ class oxArticle extends oxI18n implements ArticleInterface, oxIUrl
 
         // enabled time range check ?
         if ($this->getConfig()->getConfigParam('blUseTimeCheck')) {
-            $sDate = date('Y-m-d H:i:s', oxRegistry::get("oxUtilsDate")->getTime());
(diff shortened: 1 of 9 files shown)


floeschie:
If a product is configured to be active at 14:02 h, it must not be visible at 14:00 h. Thus, a 60 second cache time is the only acceptable option.

keywan-ghadami-oxid:
all right i will changed the description and the code to ensure that default effective cache time is maximum 60 seconds.

Shiftas:
Hi, could you rebase your branch so we could run performance tests? Also hungarian notations are no longer used and code should comply with PSR-2 code style standards.

keywan-ghadami-oxid:
Hi i removed Hungarian notations from my changes now and rebased that branch so it's compatible with the "hidden articles" features merged lately.

keywan-ghadami-oxid:
this pull request also include some utility methods, i would like to reuse for some further performance improvements in method View::getCacheLifeTime (EE Version).

PR #321, opened 2016-02-10
Updated composer.lock for case-problem fixes in testing-library

SvenBrunk:
Updated composer.lock for case-problem fixes in testing-library on windows hosts within the OXVM

### DIFF
--- a/source/composer.lock
+++ b/source/composer.lock
@@ -676,7 +676,7 @@
             },
             "dist": {
                 "type": "zip",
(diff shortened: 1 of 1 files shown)


keywan-ghadami:
@SvenBrunk thank you for that fix, it was taken directly into commit https://github.com/OXID-eSales/oxideshop_ce/commit/6d775385a3723d058281ac1c13173a50ef86791c to speed things up.
The pull request can be closed.

Shiftas:
Hi, thanks for your effort, but as Keywan mentioned we have already updated composer.lock.

PR #323, opened 2016-02-15
improve offline.html

proudcommerce:
- shop shop logo
- auto reload after 10 seconds
- maintenance message
- noindex/nofollow


marvinhagemeister:
I think it's save to use the html5 doctype and meta tags. Do you think the reload buttons are needed from an UX perspective?

proudcommerce:
> I think it's save to use the html5 doctype and meta tags.
feel free to contribute ;-) (PR or comment here)

> Do you think the reload buttons are needed from an UX perspective?
good question ...

PR #325, opened 2016-02-16
seoencoder creates entries with the wrong objectid because the object…

keywan-ghadami-oxid:
…id is not refresh.

It happened on a database with hundreds of similar urls

same-title.html
same-title-oxid.html
same-title-oxid1.html
...

And i realized the expired flag was not updated(to 0) for an seourl of some articles.

From programming point of view oxobjectid bust be set in that sql statement because it is not part of the primary key.


hkreuter:
Hi Keywan, good point, but it would be nice if you could also provide tests before we accept this pull request.

kermie:
reminder sent to @keywan-ghadami :-)

keywan-ghadami-oxid:
sorry deadlines...., but yes if you can wait some weeks, i will find time to write unit test.

keywan-ghadami-oxid:
unittest was added to this PR

PR #328, opened 2016-02-18
WIP: removed code that force a session if http request uses POST method.

keywan-ghadami-oxid:
Forcing a php session for http post requests makes trouble in several projects, because sessions may created on ajax request or form submits.
It is not gernnerall true that an POST request will need a session, the only thing "POST" does is storing data on the server and this can be independet from the current session.

On the other hand, forcing the session on POST requests may simplify other places and i am not sure what it may break.

### DIFF
--- a/source/Core/Config.php
+++ b/source/Core/Config.php
@@ -463,7 +463,7 @@ class Config extends SuperConfig
     protected function initializeShop()
     {
         $this->_processSeoCall();
-        $this->getSession()->start();
+        $this->getSession()->start(false);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Shiftas:
@keywan-ghadami - does [WIP] mean "please leave it open, more commits to expect" or "please merge ASAP, more commits to expect"?

keywan-ghadami-oxid:
WIP = Work in progress we use it in Gitlab to mark merge requests that are not ready to be merged (there are plugins for github).
The reason in this case was, that i did not test that code in the project yet. I used a workaround with skipsession Parameter in requests, and a patch to the erp plugin. 
If you want me to beta test this just leave a comment.

MantasVaitkunas:
@keywan-ghadami-oxid, yes, I'd appreciate results from your beta tests ;)

keywan-ghadami-oxid:
i did some manual tests every thing looks good (e.g. login)

keywan-ghadami-oxid:
I found some places where session is used like this:
getSession->setVariable('...','...')
without taking care if the session is open or not.
To keep caller code from beeing refactored, im thinking about to lazy start the session in get/setVariable on Post request.

PR #331, opened 2016-02-21
Fix invalid HTML in setup by upgrading Doctype to HTML5

marvinhagemeister:
The `/>` syntax is not allowed for void elements in HTML4. We should be using HTML5 anyway by now.

### DIFF
--- a/source/Setup/tpl/_header.php
+++ b/source/Setup/tpl/_header.php
@@ -22,7 +22,7 @@
 
 $editionSelector = new \OxidEsales\Eshop\Core\Edition\EditionSelector();
 ?>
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
I'd love to have this merged. Any estimate on when this will happen?

keywan-ghadami:
Hi @marvinhagemeister, 
i am sure it will be merged, html 5 doctype is easy to use, makes validation possible and does no harm
See see http://stackoverflow.com/questions/10030688/is-it-safe-to-change-to-the-html5-doctype

On the other hand i think there are no visible effects to users, are there any?
Asking because if yes we could create a bug ticket (bugs.oxid-esales.com) and this might get a higher priority.

marvinhagemeister:
@keywan-ghadami It definitely can't hurt to serve valid html markup to clients. The additional benefit is that it's much nicer for us developers to open a template without being greeted with a handful of warnings/errors.

PR #334, opened 2016-03-02
allow CamelCase for class names in metadata extends area

keywan-ghadami-oxid:
CamelCase is currently not allowed in the extends area in metatadata of modules.
If an module does it anyway it may break other modules because then the module extension chain gets corrupted. This merge request does convert the module extends keys internally into lower case so modules can not cause this kind of problems anymore.

### DIFF
--- a/source/Core/oxmodule.php
+++ b/source/Core/oxmodule.php
@@ -151,7 +151,11 @@ class oxModule extends oxSuperCfg
      */
     public function getExtensions()
     {
-        return isset($this->_aModule['extend']) ? $this->_aModule['extend'] : array();
+        $res = isset($this->_aModule['extend']) ? $this->_aModule['extend'] : array();
+        //convert class names (keys) to lower case because building extension chain is case sensitive and would fail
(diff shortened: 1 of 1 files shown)


Crease29:
+1 for this! I had some troubles with that, too a while ago.

hkreuter:
Hi Keywan, it would be nice if you could provide some tests before we accept this pull request.

Shiftas:
@keywan-ghadami-oxid - thanks for your pull request but we can't merge it for the following reasons: I don't think simply altering the notation is a good resolution for this, especially as we are shortly before introducing name spaces for modules in metadata - these would be altered as well.

kermie:
FYI: there's already a story in the backlog to allow CamelCase in metadata.php for modules in order to fix this behaviour.

keywan-ghadami-oxid:
in the bug tracker i found 2 related tickets
https://bugs.oxid-esales.com/view.php?id=3686
https://bugs.oxid-esales.com/view.php?id=5687

PR #337, opened 2016-03-04
OxNavigationTree::_loadFromFile() - fix iconv() fail leading to ommit…

jdlx:
…ed menu.xml entry …

The original regex is too greedy and will match trailing whitespaces leading to a invalid `$sCurrEncoding` which will result in `iconv()` returning `false` … thus the menu items defined will never appear. A example xml (valid!) which triggers the error:
`<?xml version="1.0" encoding="ISO-8859-15" ?>` … `$sCurrEncoding` in this case would be `ISO-8859-15" ` …

### DIFF
--- a/source/Application/Controller/Admin/oxnavigationtree.php
+++ b/source/Application/Controller/Admin/oxnavigationtree.php
@@ -116,14 +116,17 @@ class OxNavigationTree extends oxSuperCfg
         } elseif (is_readable($sMenuFile) && ($sXml = @file_get_contents($sMenuFile))) {
 
             // looking for non supported character encoding
-            if (getStr()->preg_match("/encoding\=(.*)\?\>/", $sXml, $aMatches) !== 0) {
-                if (isset($aMatches[1])) {
-                    $sCurrEncoding = trim($aMatches[1], "\"");
-                    if (!in_array(strtolower($sCurrEncoding), $this->_aSupportedExpathXmlEncodings)) {
-                        $sXml = str_replace($aMatches[1], "\"UTF-8\"", $sXml);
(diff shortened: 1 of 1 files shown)


keywan-ghadami:
I had a look at the xml specification for the encoding https://www.w3.org/TR/REC-xml/#NT-EncodingDecl it allows the encoding to be in single or doble quotes. Also the xml declaration is only allowed at the beginning but the regex scans the whole document. And the code replaces encoding name everywhere not only in the xml declaration. If i interpret the code correctly it only allows encoding only to be the last attribute. 
So +1 for this bugfix, but there is still room for improvements at this function.

jdlx:
> the regex scans the whole document

Sure? I think without the `g` flag it shouldn't be greedy … anyway, check this regex … it accounts single/double quotes … (toggle the g modifier to see the greedyness): https://regex101.com/r/pQ1uZ2/1

keywan-ghadami-oxid:
yes, it only matches once that's good, but it still scans the whole document unless the first match https://regex101.com/r/fG4gD7/1 and does not respect that encoding must not be the last attribute. maybe i will find time next week so i can help.

kermie:
@keywan-ghadami-oxid, shall we wait for your input when working on this pull request? Looks like it is not in a ready to merge state yet.

keywan-ghadami:
I am sorry i find no time for this, but my suggestions would not add much value because they are very theroeticlly so feel free to merge it now and improve it later.

MantasVaitkunas:
@jdlx, we unlikely want to merge this pull request. Can you please file a bug to the regular bug tracker: https://bugs.oxid-esales.com/ incl. all information on who to reproduce this issue so we get a more close idea on the issue? Cheers!

PR #338, opened 2016-03-06
broken utf8 in country names in setup

keywan-ghadami:
setup-country-utf8


ooxi:
How does this relate to #315 ?

keywan-ghadami-oxid:
good question. Because everything is now utf8 the data in file source/Application/Controller/Admin/shop_countries.php must also be utf8. So yes it is somehow relate to #315 because now the setup uses utf8, and data from shop_countries.php must also be utf8.

PR #339, opened 2016-03-10
Update SSL detection

revoltek-daniel:
Update SSL detection cause some Apache versions provide HTTP_HTTPS instead of HTTPS

### DIFF
--- a/source/Core/oxconfig.php
+++ b/source/Core/oxconfig.php
@@ -821,7 +821,7 @@ class oxConfig extends oxSuperCfg
     {
         $myUtilsServer = oxRegistry::get("oxUtilsServer");
         $aServerVars = $myUtilsServer->getServerVar();
-        $aHttpsServerVar = $myUtilsServer->getServerVar('HTTPS');
+        $aHttpsServerVar = ($myUtilsServer->getServerVar('HTTPS')) ?: $myUtilsServer->getServerVar('HTTP_HTTPS');
 
         $this->setIsSsl();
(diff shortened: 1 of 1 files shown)


revoltek-daniel:
Hello @keywan-ghadami-oxid ,

sorry my fault not apache, but this occurs if you use nginx as a load balancer.

The only security issue is that if the HTTPS header is not set, is that the shop could use HTTPS if there is no HTTPS needed. Cause the HTTPS header if set would preferred before the other header.

So there is only an upgrade possible but no downgrade.

keywan-ghadami-oxid:
Hello @revoltek-daniel,
if the shop would think it's https on a connection that is not HTTP, it is not directly something that would allow an attacker to get into the system, disable encryption or access data. But anyway giving an attacker one way to manipulate the system can have bad side effects, think of cookies that should be only transfered by https, or the sid parameter in the url if the protocol changes. 

Here the shop would think the connection is secure when it is not. It's like you may talk about thinks in private that you would not if you had known that it's a public communication. An attacker could use that header to make the shop "think" it has a secure connection to the user. For sure thats very theoretical and as far as i know there is no prove of concept how an attacker could use that to do something with the oxid shop, but in general i think it's a good idea to make it very hard for the bad guys.

so back to the nginx configuration i would suggest to use
in nginx config:
``` 
proxy_set_header        X-Forwarded-Proto $scheme;
```
and in Apache something like:

```
<VirtualHost *:80>
  SetEnvIf x-forwarded-proto https HTTPS=on
```

or if possible use https only

Shiftas:
Thanks for your pull request, @revoltek-daniel but we can't merge it for security reasons (see @keywan-ghadami-oxid 's comments). Also, officially, just Apache is supported (see SysReqs).

PR #324, opened 2016-02-15
Improve offline.html

marvinhagemeister:
+ Make html valid
+ Add a small maintenance note
+ Automatically attempt to redirect to index.php every 10 seconds

PR #344, opened 2016-03-30
Update offline.html

Crease29:
- Changed links from absolute to relative, because a shop isn't always installed in the root directory of a domain (e.g. /shop/)
- Moved link and style element in head element


Crease29:
Oops sorry, this should be done in #324. Will do it there.

PR #348, opened 2016-04-07
some code cleanup in oxdynimggenerator.php

keywan-ghadami-oxid:
reused $sLockName because it was
1. unused,
2. because you should not call methods more then once if you can hold the value within the same method (e.g. performance, DRY)
3. using variable names may help to document the code


MantasVaitkunas:
I have to check it, hang on

Sieg:
Thank you, Keywan!

PR #351, opened 2016-04-08
Update ArticleListController.php

ddpkts:
remove useless if statement


marvinhagemeister:
Is the if-statement really not needed? After a quick glance at the underlying functions `getBaseSeoLink` returns pretty urls whereas `getBaseStdLink` returns something like `index.php?cl=details&anid=artidà

ddpkts:
oh shit sorry, I am taking it back.
I confused getBaseSeoLink vs getBaseStdLink, thought if statement was useless, returning the same functionality ;)

PR #352, opened 2016-04-10
Javascript: Use on event handler instead of load

marvinhagemeister:
Load was deprecated long ago in jQuery 1.8 (august 2012)


Sieg:
Thank you Marvin!

PR #355, opened 2016-04-10
Use native trim() instead of jQuery's version

marvinhagemeister:
Works everywhere and includes a polyfill for IE8 and below

### DIFF
--- a/source/out/azure/src/js/IE6-8.js
+++ b/source/out/azure/src/js/IE6-8.js
@@ -89,3 +89,11 @@ if (!Array.prototype.map) {
         return A;
     };
 }
+
+// String.prototype.trim() Polyfill
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim
+if (!String.prototype.trim) {
+    String.prototype.trim = function () {
+        return this.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, '');
+    };
+}
--- a/source/out/azure/src/js/widgets/oxarticlebox.js
+++ b/source/out/azure/src/js/widgets/oxarticlebox.js
@@ -55,13 +55,13 @@
             $( ".box h3 a", oElement ).each(function() {
                 var iTitleWidth = $(this).width(),
                     iContWidth  = $(this).parent().width(),
-                    sTitleText  = $.trim($(this).text());
+                    sTitleText  = $(this).text().trim();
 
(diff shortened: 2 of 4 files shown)


Sieg:
Hi @marvinhagemeister ,

Thank you for your pull request, i want to merge it, but have some selenium tests failing. Could you please check it?

Have errors in this test files:
```
OxidEsales\Eshop\Tests\Acceptance\Frontend\NavigationFrontendTest
OxidEsales\Eshop\Tests\Acceptance\Frontend\ProductInfoFrontendTest
```

keywan-ghadami-oxid:
Hi @marvinhagemeister , @Sieg ,
whats the benefit of extracting this method from jquery? I am not a lawyer but it seems to me like taking code into the oxid from jquery that so maybe this should be doublecked for legal issues.

ooxi:
While I agree with @keywan-ghadami-oxid comment about checking for legal issues I don't think @marvinhagemeister extracted the method from jQuery. The [code](https://github.com/OXID-eSales/oxideshop_ce/pull/355/commits/2b58a24fd9ad27a4ee437a4fd15f17ab0449a855#diff-30b681c8f33a1c63dc3716a391ebff7eR93) seams to come [Mozilla's Developer Network](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim).

marvinhagemeister:
@keywan-ghadami-oxid Thanks for your interest in this PR. The first comment in the PR contains a link to the source taken from [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim). The native polyfill is far more bulletproof than [jQuery's implementation](https://github.com/jquery/jquery/blob/58c6ca9822afa42d3b40cca8edb0abe90a2bcb34/src/core.js#L319).

The reason behind this PR is simple: Remove unnecessary dependencies. You don't need jQuery for everything and at work we have a lot of complex sites which don't include jQuery at all. The trim method has been a native method for ages now (since IE 9), which it makes this PR really trivial.

I haven't had success in properly setting up the testing environment yet and I'm afraid that I can't spend more time debugging it. At my new job we don't have any OXID based shops. If someone wants to take over this PR, feel free to do so.

MantasVaitkunas:
@marvinhagemeister, I am going to close this PR. As you know, Flow theme is now the preferred choice. Also, in the meantime, the branch already conflicted. How should we proceed with this?

PR #357, opened 2016-04-20
show error messages in privatesales login form

alfredbez:


alfredbez:
fixes this bug: https://bugs.oxid-esales.com/view.php?id=6079

PR #359, opened 2016-04-22
fixed DocBlock and changed return type to boolean

alfredbez:
fixes: https://bugs.oxid-esales.com/view.php?id=6368

### DIFF
--- a/source/Application/Controller/ArticleDetailsController.php
+++ b/source/Application/Controller/ArticleDetailsController.php
@@ -962,19 +962,13 @@ class ArticleDetailsController extends \oxUBase
     }
 
     /**
-     * Template variable getter. Returns if price alarm is disabled
+     * Template variable getter. Returns if price alarm is enabled
      *
-     * @return int
+     * @return bool
      */
     public function isPriceAlarm()
     {
-        // #419 disabling price alarm if article has fixed price
-        $article = $this->getProduct();
-        if (isset($article->oxarticles__oxblfixedprice->value) && $article->oxarticles__oxblfixedprice->value) {
-            return 0;
-        }
-
-        return 1;
+        return $this->getProduct()->isPriceAlarm();
     }
 
     /**
--- a/source/Application/Component/Widget/oxwarticledetails.php
+++ b/source/Application/Component/Widget/oxwarticledetails.php
@@ -738,20 +738,13 @@ class oxwArticleDetails extends oxWidget
     }
 
     /**
-     * Template variable getter. Returns if price alarm is disabled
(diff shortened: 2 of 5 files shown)


alfredbez:
Can someone please take a look at [this](https://github.com/OXID-eSales/oxideshop_ce/blob/d653e9e0c4851612f9987b87b00b413afd4e7e45/source/Application/Model/oxvatselector.php#L84-103)? It looks very strange to me and I think it's better to return `true` instead of `0`:

```
/**
 * get vat for user of a foreign country
 *
 * @param oxUser    $oUser    given user object
 * @param oxCountry $oCountry given country object
 *
 * @return bool
 */
protected function _getForeignCountryUserVat(oxUser $oUser, oxCountry $oCountry)
{
    if ($oCountry->isInEU()) {
        if ($oUser->oxuser__oxustid->value) {
            return true;
        }

        return false;
    }

    return true;
}
```

The method is used in [oxVatSelector@getUserVat](https://github.com/OXID-eSales/oxideshop_ce/blob/d653e9e0c4851612f9987b87b00b413afd4e7e45/source/Application/Model/oxvatselector.php#L44-82), which is used in [oxVatSelector@getArticleUserVat](https://github.com/OXID-eSales/oxideshop_ce/blob/d653e9e0c4851612f9987b87b00b413afd4e7e45/source/Application/Model/oxvatselector.php#L188-202), which is used in [oxArticle@_applyVAT](https://github.com/OXID-eSales/oxideshop_ce/blob/d653e9e0c4851612f9987b87b00b413afd4e7e45/source/Application/Model/oxarticle.php#L3881-3897)

source/Application/Controller/ArticleDetailsController.php
```diff
@@ -962,19 +962,19 @@ public function getAlsoBoughtTheseProducts()
...
      *
-     * @return int
+     * @return bool
      */
     public function isPriceAlarm()
     {
         // #419 disabling price alarm if article has fixed price
         $article = $this->getProduct();
         if (isset($article->oxarticles__oxblfixedprice->value) && $article->oxarticles__oxblfixedprice->value) {
-            return 0;
```
nymo: I would return this directly...

         $article = $this->getProduct();
       return isset($article->oxarticles__oxblfixedprice->value) && $article->oxarticles__oxblfixedprice->value;

alfredbez: I think you mean:
```
return !(isset($article->oxarticles__oxblfixedprice->value) && $article->oxarticles__oxblfixedprice->value);
```

PR #361, opened 2016-04-24
method to check https only mode

keywan-ghadami-oxid:
Add a method to check if the shop runs only in https mode

### DIFF
--- a/source/Core/Config.php
+++ b/source/Core/Config.php
@@ -884,6 +884,16 @@ class Config extends SuperConfig
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Sieg:
Hi @keywan-ghadami-oxid ,

Wanted to merge the pull request, but could you please improve phpdoc with describing the return and add the test for introduced method?

keywan-ghadami-oxid:
Hi @Sieg,
I added test methods, phpdoc and rebased onto master.

PR #362, opened 2016-04-24
avoid reading session id from parameter

keywan-ghadami-oxid:
in https only mode disable session id in parameters because its not needed and not very secure.


keywan-ghadami-oxid:
Closing this PR because we need an extra config parameter for that:
blAllowSidInUrl
to allow the fallback for client with disabled cookies 
Also i need to make sure that sid i parameters are not taken into urls

PR #365, opened 2016-04-26
make adding templateblocks more fast and reliable

keywan-ghadami-oxid:
background in automated deployment on heavy load it is important that we have helper methods that add new templateblocks, with an minimum impact on the running system.

So the new code ensures that the method can be used to add template blocks without having to delete them first,
only outdated entries gets automatically deleted after new entries are added.

### DIFF
--- a/source/Core/ModuleInstaller.php
+++ b/source/Core/ModuleInstaller.php
@@ -389,46 +389,81 @@ class ModuleInstaller extends \oxSuperCfg
     /**
      * Add module templates to database.
      *
-     * @todo extract oxtplblocks query to ModuleTemplateBlockRepository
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
similar pull request https://github.com/OXID-eSales/oxideshop_ce/pull/224

Sieg:
Hello @keywan-ghadami-oxid ,

Could you please check why travis is failing for your changes?

keywan-ghadami-oxid:
Hi @Sieg  
sure i will check/adapt tests and add a method with new name.
To keep backward compatibility i will make the old method name be an alias but deprecating that.

PR #367, opened 2016-04-26
set html language regardless of performance settings

keywan-ghadami-oxid:
set the language attribute in html is even if performance option to load languages is not enabled.
To do so i use the shop internal language abbr, that should be avail without loading the setting from the database,
so we will still have performance benefit but without loosing functionality to set html language
(in most cases, in special if internal language abbr matches the ISO 639-1 language abbr).

### DIFF
--- a/source/Application/Controller/BaseController.php
+++ b/source/Application/Controller/BaseController.php
@@ -1892,23 +1892,27 @@ class BaseController extends \oxView
 
     /**
      * Returns active lang suffix
-     *
+     * usally it used in html lang attr to allow the browser to interpret the page in the right language 
+     * e.g. to support hyphons
      * @return string
      */
     public function getActiveLangAbbr()
     {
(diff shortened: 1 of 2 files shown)


Sieg:
Hello @keywan-ghadami-oxid ,

Could you please check the failing tests list in travis builds.

keywan-ghadami-oxid:
i adapted one unit tests, and rebased that feature branch now travis build is fine.

Sieg:
So the change introduces some bc break, but seems correct, because language abbreviation getting should not be dependent on performance parameter, right?

keywan-ghadami-oxid:
for me it is only a bugfix, " because language abbreviation getting should not be dependent on performance parameter," yes.

Sieg:
The related bug: https://bugs.oxid-esales.com/view.php?id=5710

PR #368, opened 2016-04-26
removed session ids from urls by default

keywan-ghadami-oxid:
it is not save anymore to transfer session id in urls.

on the other hand there are some valid usecases e.g.
 if you want to support user with disabled cookies,
 shops with extra ssl domain

but for this usecases i suggest to set
$this->blAllowSidInUrl = true in config.inc.php
and for default configuration the shop should pass the hardest security tests available

### DIFF
--- a/source/Core/Session.php
+++ b/source/Core/Session.php
@@ -508,21 +508,14 @@ class Session extends \oxSuperCfg
      */
     public function sid($blForceSid = false)
     {
-        $myConfig = $this->getConfig();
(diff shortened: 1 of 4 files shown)


keywan-ghadami-oxid:
this PR is some how related to https://bugs.oxid-esales.com/view.php?id=6389 (only for oxid members)

Sieg:
Hello @keywan-ghadami-oxid ,

Could you please check the failing tests list in travis builds.

PR #369, opened 2016-04-26
this change prevents the shop from disabling modules or going into an…

keywan-ghadami-oxid:
… unrecoverable state if modules are broken

we already have that blDoNotDisableModuleOnError configuration option, that give us some control how the shop should behave on module errors,
but in projects it has shown, that the current behavior is not a good default no matter how you set this variable.

If you set
blDoNotDisableModuleOnError to true, the shop may disable modules that you need only because of an temporary error.
and if you set blDoNotDisableModuleOnError to false, an module error may bring the shop in an unrecoverable state where you can not fix the module anymore.

So i introduced a new default that ignores errors, and tries to continue. You will still see the exception in logfiles, and maybe the broken module will may stop some areas of the shop from working,
but it's likely that will be able to fix the problem by module internals or by activating/deactivating the module in the backend.

### DIFF
--- a/source/Core/ModuleChainsGenerator.php
+++ b/source/Core/ModuleChainsGenerator.php
@@ -37,6 +37,20 @@ class ModuleChainsGenerator
     private $moduleVariablesLocator;
 
     /**
+     * @var $extensionWithEror array of moduleExtensions that can not be loaded
+     * Used within onModuleExtensionCreationError method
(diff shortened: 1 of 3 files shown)


Sieg:
Hello @keywan-ghadami-oxid ,

Could you please check the travis builds? Tests shows that something is wrong with the implementation, but the idea is good.

PR #371, opened 2016-04-27
Fix view chain

keywan-ghadami-oxid:
Remove all parent views added to the chain

### DIFF
--- a/source/Core/WidgetControl.php
+++ b/source/Core/WidgetControl.php
@@ -46,6 +46,13 @@ class WidgetControl extends \oxShopControl
     protected $_blMainTasksExecuted = true;
 
     /**
+     * Array of Views added to the view chain
+     *
+     * @var array
+     */
+    protected $parentsAdded = [];
+
+    /**
      * Main shop widget manager. Sets needed parameters and calls parent::start method.
      *
      * Session variables:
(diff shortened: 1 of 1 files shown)


Sieg:
Hello @keywan-ghadami-oxid ,

Could you please explain:

- How it was wrong before the change?
- How it will be better with the change?
- Are you using the change in actual project?

Now, what i see, you are clearing the the whole variable, maybe it could be cleared easier: Introduce new method "dropAllActiveViews".

keywan-ghadami-oxid:
@Sieg 
>How it was wrong before the change?
the issue comes when the view chain is modified, then the shop adds a new parent to the chain (which is correct), but after processing the controller (normally widgets), that parent is not removed from the chain. then we get strange errors like method not found, or sessions that are not saved. This kind of error is very hard to debug and to understand, but unlikely to happen because we do not often use parents in view chains. (maybe @robertblank remembers he had that issue with not stored sessions once in a ajax add to basket module in a subshop, and i had it once while tring to optimize varnish widgets includes to use the same parent) both where very time consuming so i decided to make this pull request.

>How it will be better with the change?
with that change we are going to cleanup afterwards, so every parent that gets added to the view chain will be removed after processing the widget.

>Are you using the change in actual project?
No sorry, i use workarounds that avoid that the view chain is modified by the shop. By using already active parents (if the parent is already active it will not be added again to view chain), and by avoiding the use of parents. But i am sorry i have no unit tests and no beta tests for this pull requests yet.

keywan-ghadami-oxid:
"Fix view chain"-branch was rebased

PR #372, opened 2016-05-02
Fix bug 6062 similar products

Alpha-Sys:
Check configured number of similar products (bug #6062), so no fatal error is thrown if the number is set to 0.


Sieg:
Thanks @Alpha-Sys , merged the last commit to our master.

PR #373, opened 2016-05-04
always fresh config object

keywan-ghadami-oxid:
having the config not as static member makes it easier to change the configuration on the fly,
for example on cron jobs i often have the use case that  i need to switch the subshop within the cronjob.
but even after refreshing the config object in registry this class still holds the reference to the old config object.
this pull request should make that easier, because each object will fetch a fresh config object.

Please do some performance test before accepting this, i personally think the change is more important then performance impact,
but to be on the save side better measure before making such decision.

### DIFF
--- a/source/Core/SuperConfig.php
+++ b/source/Core/SuperConfig.php
@@ -39,7 +39,7 @@ class SuperConfig
      *
      * @var oxconfig
      */
-    protected static $_oConfig = null;
+    protected $_oConfig = null;
 
     /**
      * oxsession instance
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
i found a related bug ticket to this topic
https://bugs.oxid-esales.com/view.php?id=4722

Sieg:
Hello @keywan-ghadami-oxid ,

The change seems good, but such changes may break currently working shop and a bunch of failing tests tells the same (Check travis builds). Could you please analize them and improve the pull request if possible?

PR #374, opened 2016-05-08
0006032: SEO Suffix, Prefix and Title doesnt support multilingual

adriankirchner:
It's as simple as deleting the call to the parent save()-method which is unnecessary because it is almost redundant, despite the fact that the edit language is not considered in the parent method.

For further information see https://bugs.oxid-esales.com/view.php?id=6032


Sieg:
Thanks @adriankirchner ,

Have just merged your pull request. Works great.

PR #376, opened 2016-05-08
Fix access to undefined variable

adriankirchner:


Sieg:
Thank you @adriankirchner .

PR #385, opened 2016-05-18
fixed annotation stating incorrect return type

SvenBrunk:
maybe you want to make sure void is always returned instead, but this way, PHPStorm will no longer complain

### DIFF
--- a/source/Core/oxseoencoder.php
+++ b/source/Core/oxseoencoder.php
@@ -709,7 +709,7 @@ class oxSeoEncoder extends oxSuperCfg
      *
      * @access protected
      *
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
Maybe we can somehow convert return value to bool, because the result from adodb should not be used in upper layers and semantic would be clear.

SvenBrunk:
If it is not meant that this value is returned, then it would be better to stick to void and instead remove the return and fix all occurences where this return value is used. (If any)

Sieg:
Thank you @SvenBrunk, i have applied the changes on our master.

PR #388, opened 2016-05-19
6327 - add own breadcrumb title for checkout step 4

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6327


Sieg:
Thank you, merged to master and ported to 5.3.

PR #390, opened 2016-05-25
6027 - Added a further SQL query part to check if the parent is buyable.

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6027


Sieg:
Thank you @michaelkeiluweit, I have just merged your pr to our master and ported to 5.3.

PR #391, opened 2016-05-25
6030 - The flag isLoaded is not set to true when to object is loaded successfully.

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6030


Sieg:
Thanks @michaelkeiluweit, just merged your pull request to our master and ported to 5.3.

PR #392, opened 2016-05-27
6171 - fixed wrong pagination behavior

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6171

The listing of the last page numbers is different to the beginning. Please have a look at the bug entry because there is a screenshot attached which shows the behavior and explains it simply. I didn't use the provided solution as the source of the problem was a few lines above.

I tested it with the theme Azure, so I can't speak for the new theme, but in Azure the box "pageHead" increased its high size, because the words "zurück" and "weiter" made the colum to wide.


Sieg:
Thank you @michaelkeiluweit, I have used the solution provided in bugtrack to fix the issue.

PR #393, opened 2016-05-27
5720 - Use the method Utils::checkUrlEndingSlash instead of a hard coded slash

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=5720


Sieg:
Thank you @michaelkeiluweit. Merged to master and ported to 5.3

PR #394, opened 2016-06-03
6083 - article sorting fields not verified against valid field from admin settings

michaelkeiluweit:
[Fixed bug 6083](https://bugs.oxid-esales.com/view.php?id=6083) 
- Added a check if the given sort field name is existing
- Replaced deprecated calls in BaseController::getUserSelectedSorting
- Wrote tests.


Sieg:
Thank you @michaelkeiluweit. The pull request was merged to our master and ported to 5.3

PR #396, opened 2016-06-03
Background for selected CMS Page categories in Navigation and in the Category tree

preethi-sivasankaran:
Category created from CMS Pages do not show the Selection as other categories. The fix will also change how the category gets displayed in the sidebar.

### DIFF
--- a/source/Application/Component/Widget/oxwcategorytree.php
+++ b/source/Application/Component/Widget/oxwcategorytree.php
@@ -72,4 +72,17 @@ class oxwCategoryTree extends oxWidget
     {
         return $this->getViewParameter("deepLevel");
     }
+
+    /**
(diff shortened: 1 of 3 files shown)


Sieg:
Hi @preethi-sivasankaran, the newly introduced method - getContentCategory is not used anywhere. Is it required?

preethi-sivasankaran:
Hi @Sieg, the new method is called in the below line:
[https://github.com/preethi-sivasankaran/oxideshop_ce/blob/64675e14dddedb7c2c2416deeac70493aacc7780/source/Application/views/azure/tpl/widget/sidebar/categorytree.tpl#L10](url)

The $oView used in this line is oxwcategorytree which does not have the method defined.

PR #398, opened 2016-06-03
Include contact information in README

ooxi:
[According to Rachel Berry](https://lwn.net/Articles/688542/), a GitHub representative, contact information should be included in a project's README file, especially if the GitHub issue tracker is disabled (which should seldomly be done)


Sieg:
Thanks @ooxi.

PR #399, opened 2016-06-05
Fix User::getSelectedAddress documentation

ooxi:
According to the documentation `User::getSelectedAddress` will return the id of the selected user address. This is not true since not the address identification but the hole address object will be returned.

Fixed documentation to reflect behaviour.


Sieg:
Thank you @ooxi.

PR #401, opened 2016-06-08
0006398: Payment type charges are not deleted changed back to a type with no costs

hendrikfreytag:
Problem are the following lines:
`$dPrice = $this->getPaymentValue($this->getBaseBasketPriceForPaymentCostCalc($oBasket));`
`if ($dPrice) {`
`...`

If there is no price it does nothing. This is no problem for the first order. But if you change it from a payment type which costs money to zero it will not remove it because of the if-statement. I added an else to explicitly set the costs to 0 if there are none.

https://bugs.oxid-esales.com/view.php?id=6398

### DIFF
--- a/source/Application/Model/Payment.php
+++ b/source/Application/Model/Payment.php
@@ -315,6 +315,11 @@ class Payment extends \oxI18n
             }
 
             $this->_oPrice = $oPrice;
+        } else {
+            // no payment costs
+            $oPrice = oxNew('oxPrice');
+            $oPrice->setPrice(0);
+            $this->_oPrice = $oPrice;
         }
 
     }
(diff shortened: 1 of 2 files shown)


keywan-ghadami-oxid:
@hendrikfreytag I removed some duplicate code in https://github.com/hendrikfreytag/oxideshop_ce/pull/1/files please consider to merge it in your branch if you agree that it is still doing the same :-)

Sieg:
Thank you @hendrikfreytag. The pull request merged and ported to 5.3 branch.

stroebjo:
I think this change also resulted to always displaying a "0,00" price for payment methods in the checkout process (azure template). Is this expected? 

I checked `application/views/azure/tpl/page/checkout/inc/payment_other.tpl` which only checks if the payment price exists, but not if it's greater 0.

PR #402, opened 2016-06-12
enable to use oxNew during utilsobject bootstrap

keywan-ghadami-oxid:
This is a minor improvement of the internal object creation process, 
that will help to simplify things by making it possible to use oxNew direcly or indirectly within the oxNew function.

technically: Set the not overloaded(by modules) version of UtilsObject as instance before creating the overloaded version.
- also some littel removement of duplicate code 
- some comments

### DIFF
--- a/source/Core/UtilsObject.php
+++ b/source/Core/UtilsObject.php
@@ -121,11 +121,12 @@ class UtilsObject
         }
 
         if (!self::$_instance instanceof UtilsObject) {
-            // allow modules
+            
             $oUtilsObject = new UtilsObject();
-
-            $classMapProvider = new ClassMapProvider(new EditionSelector());
-            $classNameProvider = new ClassNameProvider($classMapProvider->getOverridableClassMap());
+            // set the not overloaded(by modules) version early so oxnew can be used internally 
+            self::$_instance = $oUtilsObject;
+            // null for classNameProvider because it is generated in the constructor
+            $classNameProvider = null;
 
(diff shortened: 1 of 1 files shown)


ooxi:
I would oppose this change since now user code might see a partly initialized `oxUtilsObject` instance instead of a clear indication that `oxUtilsObject` is not jet available.

keywan-ghadami-oxid:
Hi ooxi thank you for feed back. Where exactly do you see the problem of making it possible to use that base instance of utilsobject during the initialisation of the extended version of utilsobject? I ask because making that possible is exactly what i want because the initialisation code then must not take care anymore how to get that instance or using new instead of oxNew. Also it fixes the chance of getting infinitive loops "max nested function" errors because oxNew will be called during initialization which will call oxNew which than again tries to initialize utilsobject and so on, i already have seen such things in projects.

But for sure the concept of making oxutilsobject and dependencies overloadable will always have the risc of that overloaded stuff is not used from the beginning so it is nothing some should without good reason anyway. I could think of flaging all objects created during the initialization so they will not be reused from instance cache or registry but that improvement could be done independent from this pull request. Also i like to reuse that basic instance in case there exists no extension for utilsobject which will be normal usecase in most projects so they should not suffer from extensibility by doing unnecessary recreation of the utilsobject instance.

ooxi:
Sorry I'm currently lacking the time for a detailed answer. Therefore ignore my objections.

PR #403, opened 2016-06-14
rename $sFile to $class

alfredbez:


Sieg:
Hi @alfredbez, We decided not to use hungarian notation for new changes anymore. Please check the http://wiki.oxidforge.org/Coding_standards page.

PR #406, opened 2016-06-14
#6411: Double slashes in request URI are replaced by single slash.

kyoya-de:


tests/unit/core/oxdynimggeneratorTest.php
```diff
@@ -36,6 +36,7 @@ class Unit_Core_oxDynImgGeneratorTest extends OxidTestCase
     public function testGetInstance()
     {
         $this->assertTrue(oxDynImgGenerator::getInstance() instanceof oxDynImgGenerator);
+        $this->assertInstanceOf('oxDynImgGenerator', oxDynImgGenerator::getInstance());
```
Sieg: Hi, is there any difference with previous line?

robertblank:
As I understand this code cleans up some mess, which is introduced in a different part.
Double slashed should not be part of $_SERVER['REQUEST_URI'] in the first place.
It would be interesting to examine (and fix) the cause of this double slashes in the URI.

@kyoya-de Could you please add some information about the context of this problem. In which circumstances do these double slashes appear? Are the produced by OXID code (should be fixed) or is the reason a certain HTTP server, reverse proxy, etc. configuration.

tests/unit/core/oxdynimggeneratorTest.php
kyoya-de: Yes. The second one is especially to check the type of an object instance.

kyoya-de:
I'm not sure about the problems source.
It happens if you use the sAltImgURL setting. You must add a tailing slash to the URL or some image URLs (for example in the wysiwyg editor) are broken. On the other hand, the tailing slash produces the double slash problem (for example at the product pictures). This mainly occurs in a multi server setup, where the OXID admin runs on the same machine together with the images. Then you've to set the alternative image URL, if the images aren't shared through all application servers.

gregorhyneck:
Was merged with commit https://github.com/OXID-eSales/oxideshop_ce/commit/15dac8bffd1ce075045f2839d3ef0b7663c97023 and backported to 5.3

PR #407, opened 2016-06-14
Exchanged all direct links to the old OXIDforge with the new links

kermie:
As OXIDforge.org was recently relaunched, some links changed with this relaunch. Removed all links to the old OXIDforge (wiki) and replaced them with the new links. Please merge as soon as possible because some people already complained that the shop installation would not be possible because some links are not valid any more. 

Pls consider to change this behaviour as well - can't be that the Shop can't be installed if a website is (temporarily) not available.


kermie:
Made changes on this post in OXIDforge so it fits again:
http://oxidforge.org/en/core-oxid-eshop-classes-must-not-be-extended.html

PR #410, opened 2016-06-16
created new abstract class Crypt

alfredbez:


alfredbez:
https://github.com/OXID-eSales/oxideshop_ce/blob/5a305e19330cc977f72db63a3e64e473a82b985b/source/Core/Utils.php#L120-136

keywan-ghadami-oxid:
but that's not used is it? https://github.com/OXID-eSales/oxideshop_ce/search?utf8=%E2%9C%93&q=strRem

alfredbez:
Ok lets remove it then

robertblank:
So would you like to suggest to deprecate the classes Encryptor and Decryptor instead?
You probably would like to adapt or close this PR then.

alfredbez:
I create another PR later

PR #411, opened 2016-06-17
refactored some methods to a more abstract one

alfredbez:


keywan-ghadami-oxid:
- added +35 Lines removed −103 Lines -> -68 Lines and the functionality stays the same :+1:
- refactored a '@internal' class: i guess that can be merged without deprecating old removed methods before.

source/Core/ModuleList.php
```diff
@@ -415,93 +415,22 @@ protected function _removeFromDisabledModulesArray($aDeletedExtIds)
...
-
-    /**
-     * Removes extension from modules events array
-     *
-     * @param array $aDeletedModule deleted extensions ID's
-     */
-    protected function _removeFromModulesEventsArray($aDeletedModule)
-    {
-        $aModuleEvents = $this->getModuleEvents();
+        $array = $this->{'getModule' . $key}();
```
robertblank: Could we introduce some "abstract" method here? Something like "getModuleConfigParametersByKey". 
I do not like the fact, that it is not clear, which methods will be called here and if this methods is defined at all. Code analysis tools like CodeSniffer and MessDetector cannot help avoiding calls to undefined methods with dynamic method names.

alfredbez:
We could even go further and mark these methods as deprecated:
```
getModulePaths
getModuleEvents
getModuleVersions
getModuleFiles
getModuleTemplates
```
And recommend to use them like so:
```php
getModuleConfigParametersByKey('Paths')
getModuleConfigParametersByKey('Events')
getModuleConfigParametersByKey('Versions')
getModuleConfigParametersByKey('Files')
getModuleConfigParametersByKey('Templates')
```

alfredbez:
Tests are passing now. I rebased it, if there is anything else you need, please, let me know.

alfredbez:
If this still **Needs improvement** tell me what to do, otherwise it would be nice if you remove that label.

alfredbez:
ping? Final commit is nearly two months old...

MantasVaitkunas:
Hello @alfredbez, we will check PR this week or in the beginning of next week. Sorry for delay.

MantasVaitkunas:
Hi @alfredbez, just merged PR and I also added some changes on top of your pull request.

alfredbez:
Hi @MantasVaitkunas thanks, I think you need to change [this](https://github.com/OXID-eSales/oxideshop_ce/blob/2c7cd1c93e9b9790239526953daf9becb8b5496f/source/Core/ModuleList.php#L131) too.

MantasVaitkunas:
@alfredbez, you mean to add constant usage?

alfredbez:
yes

MantasVaitkunas:
Done ;)

PR #413, opened 2016-06-20
refresh utilsobject modulvars on module install

keywan-ghadami-oxid:
fix ported from project, where module activation did not take affect because "moduleFiles" cache in oxUtilsObject was not refresh by moduleInstaller

### DIFF
--- a/source/Core/ModuleInstaller.php
+++ b/source/Core/ModuleInstaller.php
@@ -672,7 +672,6 @@ class ModuleInstaller extends \oxSuperCfg
         if ($this->getModuleCache()) {
             $this->getModuleCache()->resetCache();
         }
-        //clear utilsObject internal cache
-        UtilsObject::resetModuleVariables();        
+        ModuleVariablesLocator::resetModuleVariables();
     }
 }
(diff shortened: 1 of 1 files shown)


Sieg:
Hi @keywan-ghadami, is the bug: https://bugs.oxid-esales.com/view.php?id=5688 related?

keywan-ghadami-oxid:
it seams to have the same root cause, but i guess this PullRequest will not fix that issue 5688 because i did only take care about activating/deactivating modules, in my case activating was done by "ModuleInstaller" Class via oxid console.

source/Core/ModuleInstaller.php
```diff
@@ -672,5 +672,7 @@ protected function resetCache()
         if ($this->getModuleCache()) {
             $this->getModuleCache()->resetCache();
         }
+        //clear utilsObject internal cache
+        UtilsObject::resetModuleVariables();
```
robertblank: The method UtilsObject::resetModuleVariables(); is deprecated. Please use ModuleVariablesLocator::resetModuleVars instead.  See [UtilsObject.php#L382](https://github.com/OXID-eSales/oxideshop_ce/blob/64b64a29e670acd92fc2e55323df9b37ef07b64d/source/Core/UtilsObject.php#L382)

keywan-ghadami-oxid:
@michaelkeiluweit thank you for the fix, i cherry picked your commit and removed my wrong comment within your commit.
@robertblank thank you for review, please double check if there is anything left to do on this PR.

keywan-ghadami-oxid:
This is PR is not needed anymore because oxmodulecache already resets module variables by using modulevariablelocator.

PR #415, opened 2016-06-20
make clean methods availible for modules

keywan-ghadami-oxid:
cleanBillingAddress
and cleanDeliveryAddress should be available for module developers because they need that methods to secure there modules


rezonanc-oxid:
Thank you for your proposal. Unfortunately we can't accept this PR ATM because we're working on different solution from which everyone will benefit. Also we can't just change the visibility of methods on the fly.

Cheers!

PR #416, opened 2016-06-22
show offlinepage directly instead of redirects

keywan-ghadami-oxid:
the benefit of not redirecting is:
- clear status code for the caller (500) so the client knows about the error
- http standard: using 500 status with body instead of 500 with location header 
- enable the client to resend the request by refreshing the page (helps to fix the problem, can help to restore data)
- enable the client to navigate back in history

### DIFF
--- a/source/Core/Utils.php
+++ b/source/Core/Utils.php
@@ -1069,16 +1069,29 @@ class Utils extends \oxSuperCfg
     }
 
     /**
-     * Redirects to shop offline page
-     *
+     * Shows offline page
+     * @deprecated since v.5.3.0 (2016-06-28); Use Utils::showOfflinePage().
      * @param int $iHeaderCode header code, default 302
      */
     public function redirectOffline($iHeaderCode = 302)
     {
-        $sUrl = $this->getConfig()->getShopUrl() . 'offline.html';
-        $this->redirect($sUrl, false, $iHeaderCode);
+        $this->showOfflinePage();
     }
 
+    /**
+     * shows offline page
+     * directly displays the offline page to the client (browser)
+     * with a 500 status code header
+     * @return null or exit;
+     */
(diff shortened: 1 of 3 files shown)


source/Core/Utils.php
```diff
@@ -1069,17 +1069,28 @@ protected function _simpleRedirect($sUrl, $sHeaderCode)
     }
 
     /**
-     * Redirects to shop offline page
-     *
+     * Shows offline page
+     * @depricated please use showOfflinePage
```
robertblank: We have a typo here. It should be "deprecated", no "depricated".
We also need the version since when piece of code is deprecated, so the deprecation message would be something like this:
`@deprecated since v.5.3.0 (2016-06-28); Use Utils::showOfflinePage().`

source/Core/Utils.php
```diff
@@ -1069,17 +1069,28 @@ protected function _simpleRedirect($sUrl, $sHeaderCode)
...
      */
     public function redirectOffline($iHeaderCode = 302)
     {
-        $sUrl = $this->getConfig()->getShopUrl() . 'offline.html';
-        $this->redirect($sUrl, false, $iHeaderCode);
+        $this->showOfflinePage();
     }
 
     /**
+     * shows offline page
```
robertblank: Could you please add some information about the 500 header to the description of function.

keywan-ghadami-oxid:
@robertblank thank you for review, i did the changes, please double check if further improvements are needed.

MantasVaitkunas:
@keywan-ghadami-oxid with separate commit I changed deprecation message as master branch will become v6.0: https://github.com/OXID-eSales/oxideshop_ce/commit/06a227e1edaa727c796e5168e181bb8bf18b6931

PR #418, opened 2016-06-23
6141 - Added check if we really need to handle the noticelist when finalizing an order

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6141


Sieg:
Thank you @michaelkeiluweit, i have just merged your pull request to our master.

Sieg:
Ported to 5.3 branch too.

PR #421, opened 2016-06-30
Remove ldap

michaelkeiluweit:
LDAP is a enterprise feature only. The actually LDAP class isn't in the CE package, but some breadcrumbs are still there.


gregorhyneck:
We deprecated the function _ldapLogin and will remove it in near future by ourselves. A story is already planned for it.

PR #316, opened 2016-01-29
performance and encodings

keywan-ghadami-oxid:
performance improfements for utf8 mode

- do not convert utf8 encoding if already utf8
- in function _recodeLangArray
  -  removed statements without effects unset($aLangArray[$sKey]);
  -  removed second variable for array keys
- convert languange files to utf8
- change unittest to be able to detect utf8 encoding

bug fix:
 - encode utf8 to baseLang if not utf mode
 - changed unittest for that
 - change unit tests to give correct error message

PR #422, opened 2016-06-30
load language files even without 'charset' key

alfredbez:
Language files without the `charset` key are not included, e.g.
```php
<?php
$sLangName = "English";
$aLang = [
    'KEY' => 'value',
];
```

### DIFF
--- a/source/Core/Language.php
+++ b/source/Core/Language.php
@@ -920,26 +920,23 @@ class Language extends \oxSuperCfg
             $aLang = array();
             $aLangSeoReplaceChars = array();
             foreach ($aLangFiles as $sLangFile) {
-
                 if (file_exists($sLangFile) && is_readable($sLangFile)) {
                     $aSeoReplaceChars = array();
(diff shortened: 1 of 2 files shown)


keywan-ghadami-oxid:
@alfredbez what do you think about https://github.com/OXID-eSales/oxideshop_ce/pull/316 (utf-8)
do you agree that we should combine them?

keywan-ghadami-oxid:
@alfredbez i would prefer to have utf-8 to be the default encoding because shop setting does not mean that the file (that my come from a module) is encoded in the same encoding then the shop so it will break the encoding.

alfredbez:
I think #316 is good, but this and #316 are two different things, so we shouldn't combine them to one PR.
My main goal here is to load files with a missing `charset` key, that's it.
I changed the fallback encoding to UTF-8, that makes sense.

alfredbez:
I have to refactor this PR, since #316 is merged already.

PR #426, opened 2016-07-05
Fix missing DIRECTORY_SEPARATOR in mysqli driver

adriankirchner:
The mysqli driver in the current `b-5.3-ce`-branch is not working because there is a slash (DIRECTORY_SEPARATOR) missing in the `require_once`-path (file mysqli_driver.inc).


alfredbez:
https://bugs.oxid-esales.com/view.php?id=6422

Sieg:
Thank you @adriankirchner, I have just merged the pull request to 5.3 (with 4a30619).

PR #428, opened 2016-07-11
document constructor arguments support for oxNew

keywan-ghadami-oxid:


sebbbbauer:
Hi Keywan, 
good idea, but i'm not sure, if the three dots are in the phpdocs param notation (https://www.phpdoc.org/docs/latest/references/phpdoc/tags/param.html). On the other hand PHPStorm complains about 'PHPDoc for non-existing argument'.

What do you think about another solution like one or two examples?

Best regards,
Sebastian

keywan-ghadami:
Hi good point,
You can find more info here
http://stackoverflow.com/questions/13787426/how-to-declare-unlimited-parameters-in-docblock
but unless we have to support old php versions i do not know how to make
phpstorm not complaining about the missing parameter - maybe we can live
with that warning?

If not feel free to close.
Btw idea was not mine it was taken from the German dev skpe chat, because
devloper was not aware that he can pass arguments. I point this out because
providing examples like you suggested would also solve that problem.

sebbbbauer:
Hi Keywan,

fastest solution seems to accept this one, cause it will not hurt to much :)

PR #430, opened 2016-07-12
added possibility to sort the sequence of the discounts.

michaelkeiluweit:
Added functionality to custom sort the discounts sequence

It will prevent an issue which is described in the bug entry #6447 (https://bugs.oxid-esales.com/view.php?id=6447). With different databse storage engines the sequence of calculating the discount will be different. Therefore we get different total sums. Also there is no option to determine which discount should be calculated first, or to set up a whole sequence. 

The sorting works as following: The lower the sort value is, the earlier the discount will be calculated.
3 Discounts:
10%, sorting: 2
20abs, sorting: 0
20%, sorting: 1

The basket will they calculate in the sequence: 20abs, 20%, 10%.

I added the tablefield "oxsort" to the table oxdiscount. It can be accessed via the administration area, directly in the discount configuration.
It will used (sadly) in two methods:
- \Application\Model\Basket::_calcBasketDiscount
- \Core\Price::calculateDiscount

I had to write nearly the exact same code, as I didn't see an easy way to use it as a method. This is because the two places calculating the discounts in a different way (not the calculating itself but how it was coded). One time it is an little array, the other time is a full object.

The PR can be considered as a feature and as bug fix.
Feature: The shop owner is currently not capable to define which discounts should calculated in which sequence.
Bugfix: When changing the engine from MyISAM to InnoDB the total sum can be different if multiple discounts are active. With the sorting you can correct / prevent this.

For the future: One of the two listed methods should be removed and the calculation of the discounts should be in a central place.

To test it with MyISAM and InnoDb, you must open the file source/Setup/Sql/database.sql and replace the line:
`) ENGINE=MyISAM COMMENT 'Article discounts';`
with
`) ENGINE=InnoDB COMMENT 'Article discounts';`
and execute the tests again.

Without this PR some tests will fail by using InnoDB engine for the table oxdiscount.


gregorhyneck:
We implemented this feature in the meantime after the introduction of Doctrine DBAL and switch of database tables to InnoDB.

PR #432, opened 2016-07-12
Language abbreviation with underscore

preethi-sivasankaran:
The Language abbreviation that was saved with a '-' gets replaced with '_' when URI is created. But the comparison made in oxseoencoder.php, compares the uri with the Language abbreviation from the database. Hence do not match. And so repeatedly added to the URI for Languages with abbreviation using an Underscore.

### DIFF
--- a/source/Core/SeoEncoder.php
+++ b/source/Core/SeoEncoder.php
@@ -117,7 +117,7 @@ class SeoEncoder extends \oxSuperCfg
         $iDefLang = (int) $this->getConfig()->getConfigParam('iDefSeoLang');
         $aLangIds = oxRegistry::getLang()->getLanguageIds();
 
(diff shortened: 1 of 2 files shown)


sebbbbauer:
Hi Preethi,
cool, that you found that and fixed it. Could you please add a test, which covers this one? We have several unit/integrations tests for the seo things - it seems possible, that there is one, which covers nearest code.

I label this one bug, cause it seems like a bug, even there is no bug for this in the tracker.

keywan-ghadami-oxid:
https://bugs.oxid-esales.com/view.php?id=6407

preethi-sivasankaran:
@sebbbbauer 
Hi Sebastian,
I've added the unit test corresponding to the bugfix. 
Thanks.

sebbbbauer:
Already merged

PR #434, opened 2016-07-14
6450: Creation of empty fields in payments

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6450

### DIFF
--- a/source/Application/Controller/Admin/PaymentMain.php
+++ b/source/Application/Controller/Admin/PaymentMain.php
@@ -223,9 +223,10 @@ class PaymentMain extends \oxAdminDetails
             $oField = new stdClass();
             $oField->name = oxRegistry::getConfig()->getRequestParameter("sAddField");
 
-            $this->_aFieldArray[] = $oField;
-
-            $this->save();
+            if (!empty($oField->name)) {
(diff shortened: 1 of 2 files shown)


matths:
well, actually we solved the problem for ourselves as mentioned in https://bugs.oxid-esales.com/view.php?id=6450 using an overwrite construct in one of our modules:

    public function addField()
    {
        $sAddField = oxRegistry::getConfig()->getRequestParameter("sAddField");
        if ($sAddField!="") {
            return parent::addField();
        }
        return $this->save();
    }

@michaelkeiluweit your solution does only call save() when $oField->name is not empty. I think, save() should also be called when it is empty. It should just not save the "invisible" field with no name, but everything else from the form from the OXID admin. Otherwise, the user might be disappointed. Do you agree?

michaelkeiluweit:
@matths yes, you're right. 

As I see the solution could be:

```
if (!empty($oField->name)) {
    $this->_aFieldArray[] = $oField;
};
$this->save();
```

I will check this and adapt the PR when it is fine.
Maybe there is a test case missing as I wasn't alerted by executing the test. So probably I will add a test scenario, too.

Thanks for the hint! :)

PR #436, opened 2016-07-21
add semicolon to html entity

keywan-ghadami-oxid:
Html entity should/must be followed by an ;
http://stackoverflow.com/questions/18689230/why-do-html-entity-names-with-dec-255-not-require-semicolon


sebbbbauer:
Thanks for that, Keywan!

PR #439, opened 2016-08-02
Fixed invalid html

suabo:
Der invalide HTML führt dazu, dass es zu fehlern beim Speichern kommt, wenn Module den Block verwenden. Das Problem besteht ebenfalls bei allen anderen TEmplates im BE!

### DIFF
--- a/source/Application/views/admin/tpl/content_main.tpl
+++ b/source/Application/views/admin/tpl/content_main.tpl
@@ -37,169 +37,169 @@ function ShowMenueFields( iVal)
     <input type="hidden" name="cl" value="content_main">
     <input type="hidden" name="editlanguage" value="[{$editlanguage}]">
 </form>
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
Translation for my english colleges and comunity members: invalid html leads to problems if modules extending that block. Promblem exists also for all other templates in the backend.
---
@suabo The diff is large, can you help me to understand seams for me 1. the backend templates are not DRY(copy paste, and thats the reason we will need to fix it in other places as well), and what i see in your pull request is that FORM and TABLE tags where in the wrong order is it that what you fixed?

suabo:
Yes, exactly! The starting form tag is inside of the table, where the exiting tag is outside of the table. I moved the starting form element and the hidden input fields outside of the table.

alfredbez:
@keywan-ghadami-oxid Here's the diff without whitespace changes (with ?w=1): https://github.com/OXID-eSales/oxideshop_ce/pull/439/files?w=1

PR #447, opened 2016-08-09
allowing minibasket on privatesales login page

keywan-ghadami-oxid:
allowing minibasket and menu on privatesales login page so the login page so it can have the same layout then other pages.
The mini basket will of cause be empty, but it enables to reuse template structure by including the default page layout within the private sales login page:

```
[{include file="layout/page.tpl"}]
```

without this PR, developers will end up in infinitive loops because the privatesales check will cause a redirect to the login page, because the widgets included on that login page will redirect again to the loginpage.
This only a quick workaround for this bug, but i guess it will cover more then 90% of the use cases where no other widgets (with access to the user component) are used within the normal layout.

### DIFF
--- a/source/Application/Component/UserComponent.php
+++ b/source/Application/Component/UserComponent.php
@@ -94,7 +94,9 @@ class UserComponent extends \oxView
         'content',
         'account',
         'clearcookies',
-        'oxwServiceMenu',
+        //allowing some widgets so login page can use same layout as the normal shop
+        'oxwservicemenu',
+        'oxwminibasket',
     );
 
     /**


kermie:
@keywan-ghadami-oxid, to be honest, I don't get it clearly. IMO, in private sales mode, nobody should see any mini basket as the cart should be available only if you're logged in. What bug do you mean in this manner, I also couldn't find anything in the bug tracker about it? What is the number 90% about, can you underpin it somehow? Cheers!

keywan-ghadami-oxid:
@kermie on the default privatesale login page, everthing is logical, only the username and the password problem with that is that it does not look good.

The customer wants to see the same layout and design on the login page as in the shop. So from functionality the user can't use the cart or see articles, but the user has the same look and feel.

That means all elements like the minibasket must be visible, even they are not functional working.
With this Pull Request it becomes easy to modify the template for the private sales login page. Without if you try to do so in a template you will be lost in endless loops with no helpful error message.

To point it out again, this pull request does not change anything for a default shop, but brings the flexibility to make the private sales login page look like a normal login page.

PR #448, opened 2016-08-09
Switched links to the new OXIDforge.org

kermie:
See PR #407 - this is the additional PR for the maintenance branch ;)


kermie:
BTW: including test ^^

PR #454, opened 2016-08-12
fix loosing sessions with captchas in subshops

keywan-ghadami-oxid:
isCurrentUrl has again a problem that it does not correctly find the matching sub shop url
because it fails for /core/utils/verificationimg.php.

this will destroy the session of the current user because the shopid from the url will not match with that shopid from the session
and then the sessionid will be regenerated.

### DIFF
--- a/source/Core/UtilsServer.php
+++ b/source/Core/UtilsServer.php
@@ -446,7 +446,7 @@ class UtilsServer extends \oxSuperCfg
 
         //fetch the path from SCRIPT_NAME and ad it to the $sServerHost
         $sScriptName = $this->getServerVar('SCRIPT_NAME');
-        $sCurrentHost = preg_replace('/\/(modules\/[\w\/]*)?\w*\.php.*/', '', $sServerHost . $sScriptName);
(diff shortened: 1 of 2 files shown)


keywan-ghadami-oxid:
https://bugs.oxid-esales.com/view.php?id=6476

robertblank:
Hello Keywan,

I already merged the backport to v5.3.
Thanks for your work.

As of v6.0.
This behavior is currently not reproducible in v6.0 as there is not captcha any more.
Also the fixed tests pass regardless of the fixed code.
Just revert the line
`$sCurrentHost = preg_replace('/\/((?:modules|core)\/[\w\/]*)?\w*\.php.*/', '', $sServerHost . $sScriptName);`
and run the tests.
The 'core' directory is now uppercase, so in in any case it should be something like  `$sCurrentHost = preg_replace('/\/((?:modules|Core)\/[\w\/]*)?\w*\.php.*/', '', $sServerHost . $sScriptName);`
or `$sCurrentHost = preg_replace('/\/((?:modules|core)\/[\w\/]*)?\w*\.php.*/i', '', $sServerHost . $sScriptName);`
The tests would also needed to be adapted.

I will close this pull request and propose to do a major refactoring of the isCurrentUrl mechanics in a separate task.

PR #456, opened 2016-08-13
Added event binding to oxscript smarty plugin.

suabo:
As i said in dev-general-de i will add a posibility to bind js code direktly via smarty plugin oxscript.

Just do it like that:
[{oxscript add="Console.log('Bound to window.onload.');" event="load"}]
or
[{oxscript add="Console.log('Window was resized.');" event="resize"}]

### DIFF
--- a/source/Core/Smarty/Plugin/function.oxscript.php
+++ b/source/Core/Smarty/Plugin/function.oxscript.php
@@ -29,6 +29,7 @@
  * Purpose: Collect given javascript includes/calls, but include/call them at the bottom of the page.
  *
  * Add [{oxscript add="oxid.popup.load();"}] to add script call.
+ * Add [{oxscript add="$('#anylist').equalizer();" event="onload"}] to add script call.
  * Add [{oxscript include="oxid.js"}] to include local javascript file.
  * Add [{oxscript include="oxid.js?20120413"}] to include local javascript file with query string part.
  * Add [{oxscript include="http://www.oxid-esales.com/oxid.js"}] to include external javascript file.
@@ -57,7 +58,7 @@ function smarty_function_oxscript($params, &$smarty)
         }
 
         $register = oxNew('OxidEsales\Eshop\Core\ViewHelper\JavaScriptRegistrator');
-        $register->addSnippet($params['add'], $isDynamic);
+        $register->addSnippet($params['add'], $isDynamic, $params['event']);
     } elseif (isset($params['include'])) {
         if (empty($params['include'])) {
             $smarty->trigger_error("{oxscript} parameter 'include' can not be empty!");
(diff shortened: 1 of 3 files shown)


keywan-ghadami-oxid:
Hi thank you for contributing, just for my personal interest is there a reason for using native eventhandlers instead of jquery 'on' ?

And please let me ask one more question, normally oxid adds all javascripts at the end of the page
so DOM is already nearly complete when javascript gets executed, so is there any usecase where you need to register such events? 

Btw, if the core developers like the feature they will need unittest, maybe you or other community members like to add some unittests.

suabo:
I just asked me why i should implement it with jquery if the native js implementation is even as short as the one with jquery. And the native js impl. is already used in the class, so i took the native one.

There are some more events as the onload event, for example the window.onresize event can also be hooked up. If you don't exactly define when the code should be executed you never exactly know when the browser will do. To avoid some strugles with async loading it's more clean to define js code, which will manipulate or reference to dom, in the onload event.

I never looked up the tests before, is there any how-to or help page for unit tests in oxid?

keywan-ghadami-oxid:
There is some documentation in the readme of https://github.com/OXID-eSales/testing_library
I also suggest to have a look to phpunit documentation and existing tests e.g https://github.com/OXID-eSales/oxideshop_ce/blob/master/tests/Unit/Core/Smarty/PluginSmartyOxScriptTest.php

If found this discission about that topic jquery vs native  http://stackoverflow.com/questions/8996015/jquery-on-vs-javascript-addeventlistener i have no suggestion so far.

About using onload in combination with asyc javascript, might be too slow if your javascript is important for user interaction and you have other resources that may stop the browser from triggering that onload event. I am thinkimg about adding something like a "framework" attribute.

- framework="native" => native browser event
- framework="jquery" => jquery event registered with document.on(..)
- framework="oxid" that will leave the decision to oxid to have maximum performance in future oxid versions (e.g. onload could have the meaning after before closing body tag  but after asyncron javascript was loaded)

But i am still not sure how to handle javascript events within widgets when they get reloaded by ajax. Anyway i like your idea because it opens new possibilities. Let me / others know if you need help with unittests.

Sieg:
Hi @suabo, Thanks for your pull request! The feature looks interesting, but the pull request needs some improvements:
1. The oxscript usage example(doc) looks not correct, it will not work with "oxload" event, but will work with "load", or have i missed something? Additionally, you could write some basic javascript like console.log(...); in example.
2. Could you please add/improve the unit tests to test your case? Check the tests/Unit/Core/Smarty/PluginSmartyOxScriptTest.php

Sieg:
Hello @suabo, any news here?

Sieg:
Hello @suabo, as there is no progress on this question, I should close it.

It would be great if someone who is interested in this, could rebase it (preferably on latest b-6.x) and add tests, that show and test real use cases, so we could check it again and merge.

PR #457, opened 2016-08-13
Fix module url in admin

keywan-ghadami-oxid:
This fix the generation of module resource urls in admin mode, so they use the same domain in the the admin aria even when a subshop is selected(ee) and so security constraints will not be violated 
See https://bugs.oxid-esales.com/view.php?id=6474


keywan-ghadami-oxid:
There war already a related Pull request:
https://github.com/OXID-eSales/oxideshop_ce/pull/208/ 
but the solution provided there did not fix every case so this is the better fix ;-).

sebbbbauer:
Hi Michaël, hi Keywan,
thanks for the pull request - is it still work in progress?
Greetings,
Sebastian

keywan-ghadami-oxid:
Hi @sebbbbauer, Yes it's still work in progress because one customer/hoster reported that fix would not work some days ago. I am not sure how to proceed, maybe we can test it in more projects, any feedback is welcome.

nymo:
Hi @sebbbbauer @keywan-ghadami-oxid 
I did ran today in the same problem. I didn't saw this pull request before that's why I created this one: https://github.com/OXID-eSales/oxideshop_ce/pull/549
Our solution is working for us in a subshop environment but comparing with this pull request it seems I missed something? Anyway it would be nice if one of these both version could be merged into 5.3.x branch.

Sieg:
Hi @nymo, Is this version working for you? Hi @keywan-ghadami-oxid, are you planning to do something more with this pull request, or can it be analized and merged?

nymo:
Hi @Sieg, yes were still using the fix from #549 in our environment. So far any issues with it.

keywan-ghadami-oxid:
This fix works for me, the error reported in the past was caused by url rewriting in apache that was used as workaround by the hoster.

Sieg:
Hello guys, thank you for your contribution! I have just merged the pr to master and ported to 5.3

PR #459, opened 2016-08-14
use registry instead of oxNew for request object

keywan-ghadami-oxid:
Using 'new/oxnew' should be avoided because it allocates new memory,
even if oxnew may clone the object (with copy on write support) it is sometimes better to reuse objects.
For the Request object this is the case because someday you might want to give it intarnal state, it may need to parse post data from inputstream once, or you may like to manipulate that data within the controlflow of the application.


Sieg:
Thanks for your pull request, @keywan-ghadami-oxid.

PR #460, opened 2016-08-15
fix checkParamSpecialChars modifying reference

keywan-ghadami-oxid:
This is a fix for https://bugs.oxid-esales.com/view.php?id=6477


robertblank:
Hello Keywan,

Thank you for your work!
Your pull request has been merged into v6.0 and v5.3

I will close this PR now.

robertblank:
This bug cannot be fixed in v5.3
The right way to fix this bug in v6.0 is to create a separate method in v5.3 and deprecate the current method checkParamSpecialChars().
In v6.0 do not change the current method, but replace it by the new one and delete the current method checkParamSpecialChars()

PR #461, opened 2016-08-15
Alpha sys add more admin blocks

Alpha-Sys:
Add more blocks to several admin templates.


keywan-ghadami-oxid:
That is something i also need in the current project. 

Instead of doing fixing it 20 times i suggest to extract similar code into a separate template with some logic.

Alpha-Sys:
Hi Keywan,
can you give me an example for your solution (similar code in a separate template)?

alfredbez:
and again: [diff without whitespace changes](https://github.com/OXID-eSales/oxideshop_ce/pull/461/files?w=1)

PR #462, opened 2016-08-15
rethrowing the exception in order

keywan-ghadami-oxid:
rethrowing the exception after rolling back the transaction is important to give the user and the shop owner feedback.
Message must come into logfile, and must get feedback that his order was not executed.

### DIFF
--- a/source/Application/Model/Order.php
+++ b/source/Application/Model/Order.php
@@ -1368,6 +1368,8 @@ class Order extends \oxBase
         } catch (Exception $oE) {
             // if exception, rollBack everything
             oxDb::getDb()->rollbackTransaction();
+            
+            throw $oE;
         }
     }
 


hkreuter:
Hi Keywan, 
you are absolutely right, the shop must not swallow this exception. Would be great if you could add a test for this case. Where does this exception pop up, in log, in shop admin or both?

keywan-ghadami-oxid:
I guess with that PR, that exception will find it's way into the exception handler and so into the log file only.
I will take care for unittests.

Sieg:
Hi @keywan-ghadami-oxid, any news here ?

Sieg:
Hello @keywan-ghadami-oxid, as i see, this is already implemented in b38558d. Please consider reopening the PR if the case still unresolved.

PR #465, opened 2016-08-17
Added a block from version 6 to the user_main template

bertrandjackermann:
which can be extended by modules.


Sieg:
@bertrandjackermann thanks a lot for your pull request. If I understood it correctly, it is backported from v6. Unfortunately, we can't merge it into 5.3 as template changes can only be done in major steps.

PR #466, opened 2016-08-18
Performance improvement: swapping OXLANG and OXOBJECTID in index defi…

BernhardScheffold:
…nition makes index usable for all those update and delete statements without oxlang


keywan-ghadami-oxid:
great, i did a code check and found a lot of place where only oxobjectid is given in the select/delete statement so this should definitely help.

BernhardScheffold:
The shop tests took only 19.06 Minutes instead of 22.4 after this change. Of course, they do not use a real big set of articles and the run-time can vary, but still ...

[OXID-PS] Keywan Ghadami schrieb am 18.08.2016 16:35:

msornikoxid:
Many thanks to Jens Groh who has pointed us to this.

Sieg:
As Keywan mentioned, if we have a lot of queries where OXOBJECTID is used alone, this key will not help: (`OXLANG`,`OXOBJECTID`,`OXSHOPID`), but this one would help: (`OXOBJECTID`,`OXLANG`,`OXSHOPID`). Cant think of case for using a query with OXLANG but without OXOBJECTID, so the suggestion looks reasonable for me, and even not worth to put much effort on rerunning the performance.

Sieg:
Hello! Its time to do something with this proposal! So, i have checked some stuff in more detail, and it looks like we dont need to wait for complex performance testing here:

In the ShopSeo::_loadActiveUrl method (main one for routing stuff), we have the query without the lang condition:
image

For sure, we cannot add the lang here, as we want to find it out from the url, but if we could, we would get the index in:
image

With the suggested improvement, we will have the index in:
image

Am I missing something?

Sieg:
And looks like, we should move that OXLANG to the end of the index, to make it even better:
image

Sieg:
Hello guys. Have merged the changes today at 6c8f69f0f022e3155433741d8d4aea84e09cdee2. Thank you everyone.

PR #467, opened 2016-08-18
Added support für minified Resources, Added support for async loading…

suabo:
Added support für minified Resources
--
oxscript and oxstyle will now look first for minified resources in min/ folder. If there is none it will use none minified version. This gives Developers like me more freedome for building and deployment. You can now leave the grunt tasks local and use the source files instead. On a staging system a webhook can trigger the grunt task after the deployment of the repository and no paths has to be changed!
Some Devs already got some Hacks of the plugins but i think it would fit perfectly into the new releases.

Added support for async loading
--
oxscript and oxstyle will accept async="true" or async="1" to load resources async.
Async Loading of css is done by loadCSS from @scottjehl, Filament Group, Inc. (Licensed MIT)
The plugin just inserts a uglified version of it on demand.

I hope you like it!

### DIFF
--- a/source/Core/Smarty/Plugin/function.oxscript.php
+++ b/source/Core/Smarty/Plugin/function.oxscript.php
@@ -32,6 +32,7 @@
  * Add [{oxscript include="oxid.js"}] to include local javascript file.
  * Add [{oxscript include="oxid.js?20120413"}] to include local javascript file with query string part.
  * Add [{oxscript include="http://www.oxid-esales.com/oxid.js"}] to include external javascript file.
+ * Add [{oxscript include="oxid.js" async="true"}] to include local javascript async
  *
  * IMPORTANT!
  * Do not forget to add plain [{oxscript}] tag before closing body tag, to output all collected script includes and calls.
@@ -48,6 +49,7 @@ function smarty_function_oxscript($params, &$smarty)
     $priority = !empty($params['priority']) ? $params['priority'] : 3;
     $widget = !empty($params['widget']) ? $params['widget'] : '';
     $isInWidget = !empty($params['inWidget']) ? $params['inWidget'] : false;
+    $isAsync = (!empty($params['async']) && ( $params['async'] == 'true' || $params['async'] == '1')) ? true : false;
     $output = '';
 
     if (isset($params['add'])) {
@@ -65,7 +67,7 @@ function smarty_function_oxscript($params, &$smarty)
         }
 
         $register = oxNew('OxidEsales\Eshop\Core\ViewHelper\JavaScriptRegistrator');
-        $register->addFile($params['include'], $priority, $isDynamic);
+        $register->addFile($params['include'], $priority, $isDynamic, $isAsync);
     } else {
         $renderer = oxNew('OxidEsales\Eshop\Core\ViewHelper\JavaScriptRenderer');
         $output = $renderer->render($widget, $isInWidget, $isDynamic);
--- a/source/Core/Smarty/Plugin/function.oxstyle.php
+++ b/source/Core/Smarty/Plugin/function.oxstyle.php
@@ -31,6 +31,7 @@
  * Add [{oxstyle include="oxid.css"}] to include local css file.
  * Add [{oxstyle include="oxid.css?20120413"}] to include local css file with query string part.
  * Add [{oxstyle include="http://www.oxid-esales.com/oxid.css"}] to include external css file.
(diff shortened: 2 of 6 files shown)


keywan-ghadami-oxid:
Hi @suabo, really good idea to support minified files. I was wondering about the extra min folder i think it will be better to have the minified file in the same folder because it may reference other files (and minifier must handle that correctly).

About async loading of javascript, i did something like that in a module. My experiences was that it needs more features to be useful because:
- async include must be included in html header so the browser is able to load javascript asychron to html and other resources.
- all javascript that may be depended on that asynchronously loaded javascript must wait until the file is loaded.
Please correct me if i am wrong and there is a usecase without that extra features.
I found some places in code that can be optimized,
1. In the regex  dot must be escaped (/.min./ )
2. Some existing code is copy paste for async javascript

I hope my review does not sound negative, i know it's coplex topic and i hope we will see at least some parts of it in oxid 6. What do you think about separating features into independent pull requests? it may increase the chance to be accepted early by my colleagues.

suabo:
Sry for this bunch of commits. I have now removed the async support and will do another pr on this.
I have also fixed the regex and changed it to load minified resource out of the same folder.

I only thought it would be nice to have them in seperate folder to .gitignore the whole folder. This way i need to ignore all the resources which will be minified by grunt myself, which is okay. References to other files is a good point i haven't thought about.

SvenBrunk:
You are right. That is why in the setup we once had in a project, this was handled by grunt that copies/merges/minifies all files as needed to a new folder that can then be ignored (in this case, the source was outside of docroot and the target was the usual out/src/ folder)

source/Core/Smarty/Plugin/function.oxscript.php
```diff
@@ -48,6 +49,7 @@ function smarty_function_oxscript($params, &$smarty)
     $priority = !empty($params['priority']) ? $params['priority'] : 3;
     $widget = !empty($params['widget']) ? $params['widget'] : '';
     $isInWidget = !empty($params['inWidget']) ? $params['inWidget'] : false;
+    $isAsync = (!empty($params['async']) && ( $params['async'] == 'true' || $params['async'] == '1')) ? true : false;
```
Sieg: Maybe this could be simplified to something like: $isAsync = isset($params['async'])?(bool)$params['async']:false; ?

Sieg:
Hello @suabo, thanks for your pull request, but before we will check it more deeply, could you please:

* Update the description of pull request that it would represent the latest point what is done in pull request, if it doesnt.
* Rebase on latest master
* Squash commits to some logical work blocks
* Cover new code with tests

suabo:
sry i didn't answer long time. The point is debugging a minified resource is nearly impossible. You will ever see an js error in line 1, same with css. If your only do the grunt minifing magic on productiv systems you will need some ugly switches in tempalte to bind a not minified version of your resource for staging or dev systems to do proper debugging. What's your solution for that?

This merge would do the trick for this task and you don't need to care about switches for resources in a template any more. I use it in all of my projects, but have seen a lot of dirty switching for dev systems done by other devs. Maybe i just keep it to improve my workflow if you don't want to make it available to others.

SvenBrunk:
@suabo For that you usually generate map files so the debugger in the browser can un-minify the css files again (same for JS)

Sieg:
Thank you for your responses! I see this suggested functionality as a module, which can be turned on/off by needs. Consider creating one.

suabo:
Yeah it was my first intention, but i can't figure a way to do it, because the smarty plugins for oxscript an oxstyle need to be changed in order for this to work. I didn't found a solution to overwrite them.

It seems i need to take a look at these map files, maybe this PR was a stupid idea, if i knew about maps...

keywan-ghadami-oxid:
@suabo you can overwrite that plugins by
https://gist.github.com/keywan-ghadami-oxid/259b57fbfb2c70d7f146506ab0c2a38c
anyway making PR's is good thing, thank you :+1:

suabo:
Thx, for the hint keywan!

Can i really overwrite smarty plugins this, way or can i only add new ones?
I think i tried this, but wasn't able to really overwrite the oxscript and oxstyle.

But i'll give it another try, when there is some free time.

keywan-ghadami-oxid:
yes *unshift* should do the trick, so it loads your module plugins before the core plugins are loaded.

SvenBrunk:
@suabo I agree with Keywan. PRs are never a bad idea :) Even if they are closed

PR #468, opened 2016-08-19
Feature/oxmg 3 async loading resources

suabo:
## Added support for async loading

oxscript and oxstyle will accept async="true" or async="1" to load resources async.
Async loading of css is done by loadCSS from @scottjehl, Filament Group, Inc. (Licensed MIT)
The plugin just inserts a uglified version of it on demand.


suabo:
To place the async js code in the head part of the dom is in this case a template thing cause you will need to place the [{oxscript}] there. But if we put him there it would place all the registred js code there, so we need to differ here. Maybe we can place it in head with some extra parameter, like: [{oxscript async="head"}] or something.

keywan-ghadami-oxid:
@stasiukaitis-saulius i did something like this in a 5.3 + varnish project in PS. As i remember i added also some wrapper around the inline javascript to wait untli asyc js files has been loaded (by putting some kind of trigger at the end of the async and merged js). It was not that simple to cover all use cases so recommend to also review the oxps module to get a bigger picture and see if something of that module needs to be taken together with this merge request.

Sieg:
Hey guys, thanks for the input on this topic!

Yah, we are very strict on that javascript include questions in those "registrator" classes - it does not allow us to configure our javascript tags in any way, and thats the main issue probably. 

If it would be possible to save not just array of JS urls, but some more informative objects with attributes that should be added to those specific javascript includes, this specific case could be easily solved for your specific project just by adding the "async" attribute for the files you want to load differently, and requireing that async javascript from your module.

For now, there are no plans to work on this topic, and i cannot merge what is suggested here, as it does not solve the main problem, but adds complexity and problems for us to finalize and maintain it further.

I will add and mention this topic during our backlog grooming, so it will appear in the backlog. PO will decide how to handle that further.

Have a nice day!

PR #469, opened 2016-08-19
improve performance of detail page by replacing temporary variant lis…

BernhardScheffold:
…t with array.

TTFB ist improved approx. from 2 secs to 1 secs for articles with 150 variants.

### DIFF
--- a/source/Application/Model/Article.php
+++ b/source/Application/Model/Article.php
@@ -1555,7 +1555,6 @@ class Article extends \oxI18n implements ArticleInterface, \oxIUrl
                 $aVariantSelections = oxNew("oxVariantHandler")->buildVariantSelections($this->oxarticles__oxvarname->getRawValue(), $aVariants, $aFilterIds, $sActVariantId, $iLimit);
                 if (!empty($aVariants) && empty($aVariantSelections['rawselections'])) {
                     $aVariantSelections = false;
-                } else {
                 }
 
             }
--- a/source/Application/Model/Article.php
+++ b/source/Application/Model/Article.php
@@ -1578,7 +1578,7 @@ class Article extends \oxI18n implements ArticleInterface, \oxIUrl
             $this->getActiveCheckQuery($blUseCoreTable) .
             $this->getVariantsQuery($blRemoveNotOrderables, $blUseCoreTable) .
             " ORDER BY $sArticleTable.oxsort";
-        return oxDb::getDb()->getArray($sSelect);
+        return oxDb::getDb()->getAll($sSelect);
     }
    
     /**
--- a/source/Application/Model/Article.php
+++ b/source/Application/Model/Article.php
@@ -1551,20 +1551,36 @@ class Article extends \oxI18n implements ArticleInterface, \oxIUrl
         if (!isset($this->_aVariantSelections[$iLimit])) {
             $aVariantSelections = false;
             if ($this->oxarticles__oxvarcount->value) {
-                $oVariants = $this->getVariants(false);
-                $aVariantSelections = oxNew("oxVariantHandler")->buildVariantSelections($this->oxarticles__oxvarname->getRawValue(),
-                    $oVariants, $aFilterIds, $sActVariantId, $iLimit);
-
-                if (!empty($oVariants) && empty($aVariantSelections['rawselections'])) {
+                $aVariants = $this->getVariantsVarSelect();
+                $aVariantSelections = oxNew("oxVariantHandler")->buildVariantSelections($this->oxarticles__oxvarname->getRawValue(), $aVariants, $aFilterIds, $sActVariantId, $iLimit);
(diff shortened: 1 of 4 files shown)


source/Application/Model/Article.php
```diff
@@ -1551,21 +1551,37 @@ public function getVariantSelections($aFilterIds = null, $sActVariantId = null,
...
-                $oVariants = $this->getVariants(false);
-                $aVariantSelections = oxNew("oxVariantHandler")->buildVariantSelections($this->oxarticles__oxvarname->getRawValue(),
-                    $oVariants, $aFilterIds, $sActVariantId, $iLimit);
-
-                if (!empty($oVariants) && empty($aVariantSelections['rawselections'])) {
+                $aVariants = $this->getVariantsVarSelect();
+                $aVariantSelections = oxNew("oxVariantHandler")->buildVariantSelections($this->oxarticles__oxvarname->getRawValue(), $aVariants, $aFilterIds, $sActVariantId, $iLimit);
+                if (!empty($aVariants) && empty($aVariantSelections['rawselections'])) {
                     $aVariantSelections = false;
+                } else {
```
alfredbez: I think we can remove that empty block
and hello @BernhardScheffold btw :wink:

BernhardScheffold: Indeed. Thanks for the hint!

stasiukaitis-saulius:
FYI: First of all we will need to check for a performance change.
If this passes I would like to rethink a bit about the implementation.

Implementation with code from this pull request: 
- List loading is replaced with loading to array. 
- Then array is passed to another object - VariantHandler. 
- VariantHandler knows how to work with arrays and lists: _fillVariantSelections and _fillVariantSelectionsFromArray.
- VariantHandler only loads from array, other method is not called from code.

My ideas how to improve current solution:
- Maybe it is possible to optimize list loading, so it wouldn't need to change VariantHandler. 
- If not, then we would need to leave only one method. In this case we would need to deprecate the other method in legacy branch as this would be BC break.

BernhardScheffold:
There are two aspects to this performance improvement:

- reponse time is improved (a recent real example with 4000 variants has an improvment of TTFB from 10 seconds to 1 second)
- resource consumption is improved (in the 4000 variants case, the PHP memory had to be increased from 128 MB to 512 MB without the fix. Not increasing the PHP memory directly translates in the ability of having four times as many web workers on one web frontend)

The time needed for list loading is improved with each new PHP version, but not the memory consumption. Further improvements could be done by implementing assignArray in C, but this would also not improve memory consumption and would also require to add C extensions to the shop package.

I agree that _fillVariantSelections can be removed from the shop core, but I think this is the business of core development since it requires deprecation and finally elimination.

Sieg:
Possible improvements discussed:
* Fix the BC break in VariantHandler.
* _fillVariantSelectionsFromArray method is very similar to the old one, could be improved to have one and handle both cases Arrays and Objects for BC.

Sieg:
Discussed with Bernhard about the status of the PR. Keeping it open for now, he will take care of rebasing and improving it later when some free time slot will come.

BernhardScheffold:
Hi Anton,

How was this fixed now?

Regards, Bernhard

Am 9. März 2020 13:58:33 MEZ schrieb Anton Fedurtsya <notifications@github.com>:
>Closed #469.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/OXID-eSales/oxideshop_ce/pull/469#event-3110504166

-- 
Diese Nachricht wurde von meinem Gemini-Gerät mit AirMail gesendet.

Sieg:
Hi @BernhardScheffold - it was not fixed. We have discussed this PR with you a long time ago, and serious updates were required here. As there is no action on this question for 2 years already, I have decided to close it. 
If you have someone who want to do those improvements, let's reopen this pull request, otherwise - I don't see any point to have it laying open and untouched.

BernhardScheffold:
Hi Anton,

Ok, then I know the status. Since I have a module with this fix (actually it also fixes a related issue in the backend) it is not absolutely pressing but IMHO it still should be fixed in the product. I keep it in mind.

Bernhard 

Am Dienstag, 10. März 2020 schrieb Anton Fedurtsya:
> Hi @BernhardScheffold - it was not fixed. We have discussed this PR with you a long time ago, and serious updates were required here. As there is no action on this question for 2 years already, I have decided to close it. 
> If you have someone who want to do those improvements, let's reopen this pull request, otherwise - I don't see any point to have it laying open and untouched.
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/OXID-eSales/oxideshop_ce/pull/469#issuecomment-59696907

-- 
Gesendet von meinem Sailfish-Gerät.

PR #470, opened 2016-08-22
make addValidationError public

keywan-ghadami-oxid:
This class holds the "global" array of error messages, it should be possible to add error messages to the view from other classes, so we can extract & add functionality to & from other places.

### DIFF
--- a/source/Core/InputValidator.php
+++ b/source/Core/InputValidator.php
@@ -507,14 +507,28 @@ class InputValidator extends \oxSuperCfg
      *
      * @param string    $sFieldName field name
      * @param exception $oErr       exception
-     *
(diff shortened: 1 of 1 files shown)


Sieg:
Hi @keywan-ghadami-oxid, If we want to deprecate the method, we should take care about all the places where one is used. Could you please improve the pull request?

Sieg:
Sorry @keywan-ghadami-oxid, please ignore my previous comment.

Sieg:
Hello @keywan-ghadami-oxid, The changes were merged at: 689fee0409 to b-6.x. Additional steps were done on master on removing the deprecated code: 283c57dc03ac

PR #471, opened 2016-08-23
Declare var. outide the IF branch before using it in ELSE statement

bertrandjackermann:
before using it in ELSE statement


Sieg:
Its interesting why tests on travis are failing for 5.6.

kermie:
Thank you very much, @bertrandjackermann. Can you please provide a description, maybe kind of a business case, why this change is necessary? Is it mayhap something bug related?

PR #472, opened 2016-08-26
Yahoo! Finished storefront language editing!

kermie:


kermie:
Sorry, I blew it with git fetch upstream. Will send a new PR for it.

PR #475, opened 2016-08-29
restore field values after validation error

keywan-ghadami-oxid:
if register request was not accepted by the shop for any reason, the shop forgets some of the values, so the user must enter them again even if the error was related to some other field

### DIFF
--- a/source/Application/Component/UserComponent.php
+++ b/source/Application/Component/UserComponent.php
@@ -570,8 +570,14 @@ class UserComponent extends \oxView
             } else {
                 return 'register?success=1&newslettererror=4';
             }
-        } else {
-            // problems with registration ...
+        } else { // problems with registration ...
+            //restore form values for password fields
+            $this->getParent()->addTplParam('lgn_pwd',$this->getConfig()->getRequestParameter('lgn_pwd'));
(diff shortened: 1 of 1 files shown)


adriankirchner:
I guess "forgetting" the password happens on purpose to not return the given password in plain text.
Furthermore with your addition the password fields could be pre-filled eg. via redirecting the "victim" to something like `fnc=registerUser&lgn_pwd=iknowyoursecret...`
I would rather remove the variables `[{$lgn_pwd}]` and `[{$lgn_pwd2}]` in `user_account.tpl`

keywan-ghadami-oxid:
Hi Adrian, thank you for review. I will explain myself and try to make a better suggestion:

Of cause we could interpret this issue as security feature - but from my point of view this "security feature" does not add much value - passwords are transferred to the server in clear text anyway (unless ssl is used), and the server side will accept GET-Params from the URL (for any other field e.g. username).

On the other hand i find it annoying to refill form fields after submit, in case of any error.

Conclusion:
We could store the password on server side session, so it will not be returned the Browser, and the user does not need to enter the password again. And/or we could accept only POST values for the password.

keywan-ghadami-oxid:
Hi Robert thank you for the psychological aspects on this. I will think about a better solution.

PR #476, opened 2016-08-29
Update voucherserie_main.tpl

ddpkts:
Field is set as required if coupon number option is selected. That prevents generating empty vouchers.


sebbbbauer:
Hi ddpkts,
at the moment i miss a " in [this](https://github.com/OXID-eSales/oxideshop_ce/pull/476/files#diff-9e38fe1d6782a72e43d2b5b081c370f1R182) line. Can you have a look at this?
Thank you!

ddpkts:
@sebbbbauer good points, looks like missed it, added

Sieg:
Hi @ddpkts, so, by focusing in the #voucherid field, we will make it Required, and to remove that requirement, we need to click the radio button again. Maybe onfocusout should check and remove that requirement if necessary then?

Sieg:
Thanks @ddpkts, Have merged your suggested changes.

PR #481, opened 2016-09-07
shipping cost rule must include free shipping articles (please discuss)

keywan-ghadami-oxid:
The shipping cost rule must calculate the amount over all articles including free shipping articles to check if it can be applied.

A customer having a free shipping article and a normal article in basket but the total amount of both articles should make the free shipping cost rule to be applied.

Without this pull request only normal articles are included in the amount and so the free shipping does not apply even if the total amount would be higher than the minimum for the free shipping.

There was a discussen on skye chat with no final result if this is a bug or feature.
So i kindly ask all readers to participate and ask questions and give feedback.
If the old behavior is really an feature we could make it configurable.

### DIFF
--- a/source/Application/Model/Delivery.php
+++ b/source/Application/Model/Delivery.php
@@ -234,36 +234,35 @@ class Delivery extends \oxI18n
         } else {
 
             $this->_blFreeShipping = false;
-
-            switch ($this->getConditionType()) {
-                case self::CONDITION_TYPE_PRICE: // price
-                    if ($this->getCalculationRule() == self::CALCULATION_RULE_FOR_EACH_PRODUCT) {
-                        $dAmount += $oProduct->getPrice()->getPrice();
-                    } else {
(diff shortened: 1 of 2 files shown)


keywan-ghadami-oxid:
discussed again in german skype chat https://join.skype.com/gsbk3FTaHuar
and it is only mentioned as feature because the documentation can be interpreted in that way.
Because it is written that this articles are ignored for shipping cost rules, but better formulation would be that for this articles will no shipping cost be added.
https://www.oxid-esales.com/de/support-services/dokumentation-und-hilfe/oxid-eshop/einrichtung/artikel/registerkarte-erweitert.html

AMartinNo1:
I would like to see this as a configurable option. So every shop admin can decide what he/she wants to use.

leofonic:
This was probably changed here: https://bugs.oxid-esales.com/view.php?id=1980 because free shipping articles might have a weight for example and if this weight is added to total weight and shipping costs are calculated by weight then the customer would pay extra shipping costs for the article although it is marked free shipping.

keywan-ghadami-oxid:
@leofonic thank you very much for pointing this out. It's clear now that we can't fix ticket 6499 without a new configuration option (maybe on the delivery cost rule itself).

PR #482, opened 2016-09-07
Category tree, selected category search fix

Lezas:
About error, in category list wrong category was selected. This was because parent category ID was '003' and other category ID was '03'. And then '==' comparison tries to compare these strings, it interprets them like a numbers. Fixed by strcmp() which checks strings in binary level.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,22 +21,22 @@
 
 # Theme
 /source/Application/views/*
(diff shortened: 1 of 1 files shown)


robertblank:
Relates to https://bugs.oxid-esales.com/view.php?id=6500

PR #483, opened 2016-09-13
remove copypaste frameset in admin area

keywan-ghadami-oxid:
Removing duplicate code because it is the root of all evil.
- extracted duplicate code into an other template and included that.
- removes about 80% code from the affected files.
- it could be done better by using only one template at all, but i suggest refactoring on template level with 'include' statements first because you can use that pattern almost everywhere without any BC breaks and with low efforts.


MantasVaitkunas:
@keywan-ghadami-oxid, for next time: could you please try to open web page and check if the place you are improving is not broken. Because after merge fatal error occurred in some of admin pages.

keywan-ghadami-oxid:
sorry for that - i will try to find a solution for that - (working from android/github-web-client most of the time)

PR #485, opened 2016-09-14
removed changes from PR390 therefore unbuyable parents can be assigned again.

michaelkeiluweit:
Please see https://bugs.oxid-esales.com/view.php?id=6501#c11770 for more information, especially 

> Reporter in 0006027 says it is wrong that you can assign an unbuyable item as a discount. He does not say it is wrong to assign an unbuyable parent to a discount. It is a good thing if unbuyable parents can be assigned, just as you can assign categories, which are also not buyable.


robertblank:
The bug entry describes, why the code was only partially reverted

PR #487, opened 2016-09-22
Add the ability to use all configured columns for sorting

keywan-ghadami-oxid:
For default shop is perfectly find to accept only sorting request parameter that do exists on the article table,
but for EE shops the search often does not only go against the article table.


keywan-ghadami-oxid:
That patch also fixes an other bug. As the article object supports lazy loading it is not guaranteed that it nows all available fields. Thats the reason somtimes after a cache clear the sorting request parameter was not accepted.

stasiukaitis-saulius:
Hi Keywan,

Similar change was introduced to both master and 5.3 branches. Hope it will work for you.

https://github.com/OXID-eSales/oxideshop_ce/commit/ff2098a5143b6556eecd6418caf6924ca970aaf0
https://github.com/OXID-eSales/oxideshop_ce/commit/4dd85153f3d1ccc1fe901dd51d785a60f42162a6

Cheers

keywan-ghadami-oxid:
yes that is what I needed :-)

PR #488, opened 2016-09-23
fix field length for oxwidth and oxheight

Alpha-Sys:


keywan-ghadami-oxid:
Hi Fabian, thank you very much for you contribution. I am doing the first code review:

This is a bug fix for the admin backend in extended tab for the products. In two fields the maximum size of the fields was not set correctly. 
Only two changes in a template so it is sure that it will have NO side effect and NO bc breaks.

PR #489, opened 2016-09-23
add active column to manufacturer list

Alpha-Sys:


keywan-ghadami-oxid:
Hi Fabian thank you for this. I am writing some results of my code review:
- adds active column to manufacturer list in admin backend 
- only template changes to view the active flag
- no bc break

source/Application/views/admin/tpl/manufacturer_list.tpl
```diff
@@ -76,6 +81,7 @@ window.onload = function ()
             [{if $listitem->getId() == $oxid}]
                 [{assign var="listclass" value=listitem4}]
             [{/if}]
+            <td valign="top" class="[{$listclass}][{if $listitem->oxmanufacturers__oxactive->value == 1}] active[{/if}]" height="15"><div class="listitemfloating">&nbsp</a></div></td>
```
keywan-ghadami-oxid: I found a minor copy paste bug: a missing semicolon at the end of the entity &nbsp.
As far i know this is deprecated since html5 but will still work in most browsers because it is followed by a special character.

PR #495, opened 2016-09-30
vat message for b2b users with valid ustid

keywan-ghadami-oxid:
vat messages say vat is included in the price but that's wrong
https://bugs.oxid-esales.com/view.php?id=6516

### DIFF
--- a/source/Application/Controller/BaseController.php
+++ b/source/Application/Controller/BaseController.php
@@ -2903,8 +2903,9 @@ class BaseController extends \oxView
 
         $country = oxNew('oxCountry');
         $country->load($user->getActiveCountry());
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
just realized that there might be a better way to fix this by using getUserVat from VatSelector class
https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Application/Model/VatSelector.php#L59

Sieg:
Hello @keywan-ghadami-oxid, thanks for your pull request. Are you planning to make further improvements on this pull request?

Sieg:
Hello @keywan-ghadami-oxid, some points to update the pull request:
* Cant we use getFieldData method to get values?
* How about strict comparison for values (===) so checking for null will be not needed.

Sieg:
Thank you for your pull request. I have merged it at fa16a934e74a3425a3261c514f7176a707759349. Have a nice day.

PR #498, opened 2016-10-11
Fix wrong calculation in rss

dassaffe:
Added a check if the parent article is buyable.
If not, the varMinPrice is taken instead of the price of the parent article.

### DIFF
--- a/source/Application/Model/RssFeed.php
+++ b/source/Application/Model/RssFeed.php
@@ -238,10 +238,19 @@ class RssFeed extends \OxidEsales\Eshop\Core\Base
             $oItem = new stdClass();
             $oActCur = $this->getConfig()->getActShopCurrencyObject();
             $sPrice = '';
-            if ($oPrice = $oArticle->getPrice()) {
-                $sFrom = ($oArticle->isRangePrice()) ? Registry::getLang()->translateString('PRICE_FROM')." " : '';
-                $sPrice .= ' ' . $sFrom . $oLang->formatCurrency($oPrice->getBruttoPrice(), $oActCur) . " " . $oActCur->sign;
+
+            // check if article is a variant
+            if ($oArticle->isParentNotBuyable()) {
+                $oPrice = $oArticle->getVarMinPrice();
+            } else {
+                $oPrice = $oArticle->getPrice();
             }
+
+            if ($oPrice) {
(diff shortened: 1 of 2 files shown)


keywan-ghadami-oxid:
Please double check if the the Pull Request is necessary according to the comment in 
https://bugs.oxid-esales.com/view.php?id=6522
the problem was not reproducible, so it might be the ticket needs more information or this pull request is normally not needed.

dassaffe:
Please double-check the steps to reproduce the issue. It is reproducable, as explained in the ticket. This error occures even in the current demoshop (without setting up any kind of options). 

http://demoshop.oxid-esales.com/professional-edition/startseite/
in the FE the price for Kuyichi Jeans ANNA is "from 92,90", in the RSS it is set to "from 99,90". 

Thanks for doublechecking

keywan-ghadami-oxid:
Yes you are right, i was also able to reproduce. Thank you for the pull request and clarifications.

stasiukaitis-saulius:
Hi Daniel @dassaffe,

Thank you for this pull request. I would like to accept it, but before doing this:
- Can you write a test which would fail before test and would work with your code. 
- Write ID of the bug to a comment above new test.
Something like it is done there https://github.com/OXID-eSales/oxideshop_ce/blob/master/tests/Unit/Application/Model/ArticleTest.php#L2062

dassaffe:
Sorry for the delay, have still some stuff to do, i'll add them asap

alfredbez:
I've added the unit-tests. You can cherry-pick them and see, that they're failing without the fix provided by @dassaffe 

@dassaffe Go ahead and take a closer look at the tests, try to understand what I've done and ping me, if you have questions :wink:
level :up: by contributing to OpenSource

alfredbez:
> * Can you write a test which would fail before test and would work with your code.
> * Write ID of the bug to a comment above new test.

done :white_check_mark:

alfredbez:
Is there still something missing?

You can see a screenshot which shows the error in this blog-post: https://www.web-grips.de/blog/oxid-rss-feed-fehlerhafte-preisberechnung-beheben/ (german)

Is there any chance to get this backported since this is an actual bug?

Sieg:
Hello guys, its been a while for this one ;) Thanks for your input, i have checked and merged it to b-6.1.x recently, so the fix should be available in next patch. Have a nice day there!

PR #501, opened 2016-10-26
feature: Store the price without discounts

keywan-ghadami-oxid:
What:
This pull requests adds the ability to retrieve the price without discounts.  

Why:
This is helpful in projects if you need to display the original price next to the effective price.
As the price logic with vats and discounts can become complicate, without this methods developer tending to create even more logic on top that in there module, making thing unmaintainable.

BC Break: 
no

Side effects:
if discounts are in effect it uses some bits additionally memory to store the original price, but that is nothing that should be measurable in practical use.

### DIFF
--- a/source/Core/Price.php
+++ b/source/Core/Price.php
@@ -60,7 +60,14 @@ class Price
      */
     protected $_aDiscounts = null;
 
-
+    
+    /**
+     * The price without discounts
+     * @var \oxPrice
+     */
+    protected $priceWithoutDiscounts = null;
+    
+    
     /**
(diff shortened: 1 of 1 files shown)


Sieg:
Hi @keywan-ghadami-oxid, thanks for your pull request. Looks good, but could you add a test for your case?

Sieg:
Hey @keywan-ghadami-oxid, we had some discussion about this one. From the first glance it looked fine, but seems its really easy to break this object by running some of the methods several times. We have decided to not take this functionality. Have a nice day.

PR #503, opened 2016-11-07
added commented out $this->blSkipViewUsage to config.inc.php

vanilla-thunder:
People having problems with views will just have to uncomment this line instead of searching the internet. It also prevents copy-paste and layer 8 errors.


Sieg:
Hello, thanks for the pull request. The line already present in our .dist file, question is not viable anymore.

PR #504, opened 2016-11-07
remove duplicate if-statement

alfredbez:


rezonanc-oxid:
Yes, it was duplicate indeed! Thanks!

PR #506, opened 2016-11-10
6498: external link on categories are rendered with aditional questionmark

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6498


keywan-ghadami-oxid:
same topic as
https://github.com/OXID-eSales/oxideshop_ce/pull/480

michaelkeiluweit:
Thanks for pointing that out. I changed my comment in the bugtracker, too: https://bugs.oxid-esales.com/view.php?id=6498#c11862

PR #508, opened 2016-11-15
Fix php warning

GM-Alex:
Fix the Invalid argument supplied for foreach() warning.


rezonanc-oxid:
Simple yet effective, thanks!

PR #509, opened 2016-11-16
remove unused parameter

alfredbez:


keywan-ghadami-oxid:
Hi Alfred,
thanks you for contribution:
I guess we can not be sure that this parameter is not used in modules, can we?.
btw it is called within:
https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Application/Model/Basket.php#L1864

PR #510, opened 2016-11-16
performance reuse oxlocater result

keywan-ghadami-oxid:
improves performance about 2% for detail page, by not calling oxlocater again in the the widget.

### DIFF
--- a/source/Application/Component/Widget/ArticleDetails.php
+++ b/source/Application/Component/Widget/ArticleDetails.php
@@ -929,25 +929,30 @@ class ArticleDetails extends \OxidEsales\Eshop\Application\Component\Widget\Widg
 
         parent::render();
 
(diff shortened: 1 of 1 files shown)


stasiukaitis-saulius:
Hi @keywan-ghadami-oxid ,

Did you check if it works with Varnish?

keywan-ghadami-oxid:
No, sorry i can't i did it with 5.3.1 in a project without varnish, but it's a good point. Maybe someone else reading this find some time to test it with varnish. I guess it will do no harm there but performance will also not be affected together with varnish. Btw i thought ci could run selenium tests with varnish doesn't it?

stasiukaitis-saulius:
Sure, just wanted to mention that widgets act a bit different with Varnish turned on. 
We'll check this pull request in CI before accepting it and it will show if there is a regression with Varnish.

PR #511, opened 2016-11-17
Made some changes like fixed typos, moved similar parameters together…

michaelkeiluweit:
I did a few things here, as the time is perfect:

1. Moved all lines to the left. The file will be included into a PHP file, but it isn't necessary for a human nor the machine to have those whitespaces on each line.

2. Changed multiline comments to inline comments:
Example:
```
//aaa
//aaa
$this->a = null;
```
to
```
/*
 * aaa
 * aaa
 */
$this->a = null;
```
3. Moved parameters because they were in the "Enterprise Edition only" section.
- blShowUpdateViews
- blDelSetupDir

4. Removed as it isn't used anymore. #5337
`$this->iDebugSlowQueryTime`

5. Removed @set_time_limit(3000) as this should be configured in the php.ini file and not in the shop code.
```
// If default 30 seconds is not enougth
// @set_time_limit(3000);
```
6. Added the configuration parameter `$this->blSeoMode = true;`

7. Normalized the arrays.

8. Fixed typos, changed some beginnings to uppercase.

There are two questions left: 
- Should all lines have an inline comment to have some consistency? It wouldn't be PSR like, but it would be a more clear form.
- Variables with no value like arrays are sometimes initiated with array() and sometimes will null. Should it be changed to array()? Same for strings. Sometimes it's '' and sometimes null. For the arrays I checked the unit tests and it seems like it wouldn't be a problem to change the value to array().

### DIFF
--- a/source/config.inc.php.dist
+++ b/source/config.inc.php.dist
@@ -20,191 +20,224 @@
  * @version   OXID eShop CE
  */
 
-    /** database connection information */
-    $this->dbType = 'pdo_mysql';
-    $this->dbHost = '<dbHost>'; // database host name
-    $this->dbPort  = 3306; // tcp port to which the database is bound
-    $this->dbName = '<dbName>'; // database name
-    $this->dbUser = '<dbUser>'; // database user name
-    $this->dbPwd  = '<dbPwd>'; // database user password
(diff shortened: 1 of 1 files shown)


michaelkeiluweit:
see #524

PR #516, opened 2016-12-01
feedback on firewall blocking license check

keywan-ghadami-oxid:
make it possible to use admin backend, and see the error message if firewall is blocking the license check.
without this fix it is not possible to review or change the current license if the shop runs behind a firewall.
screenshot

### DIFF
--- a/source/Application/Controller/Admin/ShopLicense.php
+++ b/source/Application/Controller/Admin/ShopLicense.php
@@ -8,6 +8,7 @@ namespace OxidEsales\EshopCommunity\Application\Controller\Admin;
 
 use oxRegistry;
 use oxSystemComponentException;
+use OxidEsales\Eshop\Core\Exception\ConnectionException;
 
 /**
  * Admin shop license setting manager.
@@ -101,8 +102,11 @@ class ShopLicense extends \OxidEsales\Eshop\Application\Controller\Admin\ShopCon
         $oCurl->setMethod("POST");
         $oCurl->setUrl($sUrl . "/" . $sLang);
         $oCurl->setParameters($aParams);
-        $sOutput = $oCurl->execute();
-
+        try {
+            $sOutput = $oCurl->execute();
+        } catch (ConnectionException $ex) {
+            $sOutput = "Connecting to $sUrl failed, please check network and firewalls";
+        }
         $sOutput = strip_tags($sOutput, "<br>, <b>");
         $aResult = explode("<br>", $sOutput);
         if (strstr($aResult[5], "update")) {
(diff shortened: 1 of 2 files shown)


source/Application/Controller/Admin/ShopLicense.php
```diff
@@ -117,8 +117,11 @@ protected function _fetchCurVersionInfo($sUrl)
         $oCurl->setMethod("POST");
         $oCurl->setUrl($sUrl . "/" . $sLang);
         $oCurl->setParameters($aParams);
-        $sOutput = $oCurl->execute();
-
+        try {
+            $sOutput = $oCurl->execute();
+        } catch (oxException $ex) {
```
SvenBrunk: Isn't that a bit to general? Is there a more specific exception for this case?

keywan-ghadami-oxid: no sadly it throws an basic "oxException" on curl errors. See
https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/Curl.php#L314

I guess it would be possible to create a "CurlException" in this place without any BC break. I would like to see that happen but i will not include it in this pull request to keep things atomic.

Sieg: What is the error message of exception? maybe that message should be used? Maybe the resource is just not available, not because of firewall?

Sieg:
@keywan-ghadami-oxid What is the error message of exception? maybe that message should be used? Maybe the resource is just not available, not because of firewall?

keywan-ghadami-oxid:
I improve this by using the more specific "ConnectionException" and more generic error message, see screenshot.

source/Application/Controller/Admin/ShopLicense.php
```diff
@@ -101,8 +102,11 @@ protected function _fetchCurVersionInfo($sUrl)
         $oCurl->setMethod("POST");
         $oCurl->setUrl($sUrl . "/" . $sLang);
         $oCurl->setParameters($aParams);
-        $sOutput = $oCurl->execute();
-
+        try {
+            $sOutput = $oCurl->execute();
+        } catch (ConnectionException $ex) {
+            $sOutput = "Connecting to $sUrl failed, please check network and firewalls";
```
alfredbez:
```suggestion
            $sOutput = "<div class='errorbox'>Connecting to $sUrl failed, please check network and firewalls</div>";
```

Would be even better to extraxt the message to a language-string like this:
```suggestion
            $sOutput = '<div class="errorbox">';
            $sOutput .= sprintf($oLang->translateString('LICENCE_CHECK_FAILED', $iLang, true), $sUrl);
            $sOutput .= '</div>';
```

alfredbez: image

Sieg:
Hey guys, looks like the PR is not viable anymore:
image

Fixed here: https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Application/Controller/Admin/ShopLicense.php#L94

PR #517, opened 2016-12-05
added oxuser__oxusername to the list of required fields

vanilla-thunder:


vanilla-thunder:
Actually i changed only aMustFillFields oxconfig valuein row 55, but github must have some troubles with file encoding, since it uses utf-8 by default. I didn't change the other 78 rows showed in diff

vanilla-thunder:
Adding oxuser__oxusername to the list of required fields for user registration makes it impossible for existing customers to pass checkout step 2, since oxid seems to search for all required fields and billing address form does not contain oxuser__oxusername

PR #519, opened 2016-12-06
cast value to float

alfredbez:
fix failing php7.1 build

see also #518 and eb6af3562e60095c48f3267a1a9fb4b72cb36490

btw: We use `(float)` and `(double)` in the code. Maybe we should use just one of these?


rezonanc-oxid:
Sorry, we had a meeting today, so I was not able to introduce this change yet, but it's good that someone keeps an eye on these things! Thanks again!

PR #523, opened 2016-12-21
Updates tpl & ajax controller to allow custom fields in table oxobjec…

AMartinNo1:
Updates tpl & ajax controller to allow custom fields in table oxobject2attribute. In order to use this each editable field must have following attributes: "id" and "data-param-id". The latter is equal to the key of the object stored in the js variable `oParam._oData`.

oxideshop_ce-feature-customize-oxobject2attribute

### DIFF
--- a/source/Application/Controller/Admin/ArticleAttributeAjax.php
+++ b/source/Application/Controller/Admin/ArticleAttributeAjax.php
@@ -156,10 +156,13 @@ class ArticleAttributeAjax extends \ajaxListComponent
                 $objectToAttribute->init("oxobject2attribute");
                 if ($objectToAttribute->assignRecord($select)) {
                     $objectToAttribute->oxobject2attribute__oxvalue->setValue($attributeValue);
-                    foreach ($attributeCustomData as $key=>$value)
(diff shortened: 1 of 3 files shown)


AMartinNo1:
> **keywan-ghadami-oxid:**
> In ArticleAttributeAjax.php line 163: using a request value as (part of) a field name (without a whitelist) is dangerous, it might open a door for attackers because they can inject any fieldname (exisiting or even invalid fieldnames). I know the backend is password protected but i think new code should try to avoid such things to keep the risc of security issues low.

I see. Good aspect. I'll add a respective class attribute like `ArticleAttributeAjax::$editableFields`. In order enhance the implementation I would like to modify the ajax template by a foreach that iterates through `ArticleAttributeAjax::$editableFields` to generate the fields accordingly. However, therefor I need some support regarding UnitTests and TPLs. Let's say in the template the editable fields are generated based on the above variable. Would it affect how we write the UnitTest? If so, in what manner?

AMartinNo1:
This is pretty old so, no tests provided etc. Will close it.

PR #524, opened 2016-12-22
normalized the comments, fixed some typos.

michaelkeiluweit:
I did a few things here with the background that the file is basically a file which represents a part of class attributes from \OxidEsales\EshopCommunity\Core\Config.

1. Normalized the comments. 
 - A comment which needs only one line is beginning with two Slashes and a whitespace.
 - Multiple line comments have the comment pattern of a method comment.
2. Fixed typos.
3. Added or removed line breaks.

What is missing and will not be added are the public / protected / private keywords for the class attributes as this could have to much impact.

### DIFF
--- a/source/config.inc.php.dist
+++ b/source/config.inc.php.dist
@@ -20,7 +20,7 @@
  * @version   OXID eShop CE
  */
 
-    /** database connection information */
+    // Database connection information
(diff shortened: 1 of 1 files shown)


rezonanc-oxid:
Thanks! One small note, please tend to keep a new line at the EOF.

PR #525, opened 2016-12-23
This is the next step after normalizing the comments (see PR #524)...

michaelkeiluweit:
This PR is an addition to the PR #524  
  
Seperated the database, url, directory parameters to own blocks.
  
Grouped   
- Enterprise Edition parameters together und moved "not Enterprise" parameters above.  
- Parameters together which have certain similarity.  
  
Removed   
- $iDebugSlowQueryTime as it isn't anywhere used  
- @set_time_limit(3000); as this should be defined directly in the php.ini file and not in the shop.  
  
Restructered the arrays and replaced double quote with single quotes.


rezonanc-oxid:
Few notes from my side:

* Not sure why this was changed "// database connection information" I think it would be better to keep all first letters as capital ones.
* "$this->sCompileDir  = '<sCompileDir>';" I would suggest to add a comment to explain what 'compile' means in this context.
* "@set_time_limit(3000); as this should be defined directly in the php.ini file and not in the shop." Could you provide more arguments to this? ATM this helps with the long running scripts that we have.
* I would suggest to use the [] array definitions.

Would you be willing to make the changes? In case not, I'll just merge this and apply changed on top of this.

PR #526, opened 2016-12-23
Removed set_default_timezone from config.ini.php.dist

michaelkeiluweit:
This is PR is on top of the PR #525. I removed the php setting function set_default_timezone as this should be set directly in the php.ini file and not in the shop.

### DIFF
--- a/source/config.inc.php.dist
+++ b/source/config.inc.php.dist
@@ -20,18 +20,25 @@
  * @version   OXID eShop CE
  */
 
-    // Database connection information
+    // database connection information
     $this->dbType = 'pdo_mysql';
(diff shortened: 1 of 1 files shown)


rezonanc-oxid:
This PR contains commit of other PR (#525). Please avoid this as much as possible, because this PR now has a dependency on other PR, i.e. if the dependent PR will be rejected, this will be rejected as well.

rezonanc-oxid:
Travis is also showing that our tests is coupled with this setting. So I guess it's safe to close this one, would you agree? Or would you be willing to improve the tests as well?

michaelkeiluweit:
I will make a new PR with adjusted tests. But it might take awhile so I will close this one for now.

PR #527, opened 2016-12-23
Removed first four whitespaces

michaelkeiluweit:
This is PR is on top of the PR #525. It basically removes the first for whitespaces and does nothing more.

Be aware that the changing of the PR #526 is not in this PR. As both (#526 and this one) are branches of #525.

### DIFF
--- a/source/config.inc.php.dist
+++ b/source/config.inc.php.dist
@@ -20,18 +20,25 @@
  * @version   OXID eShop CE
  */
 
-    // Database connection information
+    // database connection information
     $this->dbType = 'pdo_mysql';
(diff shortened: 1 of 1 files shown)


rezonanc-oxid:
I think it would be better (time to accept PR wise) if this kind of commit would be included after the mentioned PRs would be merged to master. Please avoid PR dependency...

Sieg:
Thanks, i have just applied this fix on b-6.x on https://github.com/OXID-eSales/oxideshop_ce/commit/2ec3d6eb29db4a01ba890db23b9df842f3ffe2d8, set you as the author.

PR #533, opened 2017-01-20
removed span from lang file fixes 0006546

nymo:
see https://bugs.oxid-esales.com/view.php?id=6545


ambrosiustopor:
Thanks for your contribution.
This issue has been fixed.

PR #534, opened 2017-01-25
New block in admin, before the list of installed modules

mparienti:
I need a new block in the modules list page.

This will allow us to filter extended classes.

Thanks in advance


Sieg:
Hi @mparienti, I have added the new block on your desired location, but renamed it a little. Please take a look!

PR #536, opened 2017-01-30
make oxcurl compatible with proxy settings

keywan-ghadami-oxid:
This is important for customers running oxid in secure environments where the shop can only communicate through proxies.
With this commit it is possible to setup a default configuration by the hoster e.g.:
```
   $stream_default_opts = array(
    'ftp' => array(
        'proxy' => "tcp://127.0.0.1:8888",
        'request_fulluri'=>true
    ),
    'http'=>array(
        'proxy'=>"tcp://127.0.0.1:8888",
        'request_fulluri' => true
    ),
    'https'=>array(
        'proxy'=>"tcp://127.0.0.1:8888",
        'request_fulluri' => true
    )
   );
stream_context_set_default($stream_default_opts);
```
and things like license check will work again

### DIFF
--- a/source/Core/Curl.php
+++ b/source/Core/Curl.php
@@ -385,7 +385,13 @@ class Curl
         if (!is_null($this->getHeader())) {
             $this->_setOpt(CURLOPT_HTTPHEADER, $this->getHeader());
         }
-        $this->_setOpt(CURLOPT_URL, $this->getUrl());
+        $url = $this->getUrl();
+        $this->_setOpt(CURLOPT_URL, $url);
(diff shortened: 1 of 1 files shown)


Sieg:
Hello @keywan-ghadami-oxid, how about making this configurable through the admin? A little note more - parse_url have second parameter possible, probably it would be more clear to use it.

keywan-ghadami-oxid:
yes I agree in both points, I think that line can be simplified and also admin setting would be better. I closing this in favor of publishing a module fir this use case.

Sieg:
Yes, probably having a module for this case is the right decision, as its quite specific.

PR #537, opened 2017-02-07
User::delete method must return boolean, but there is a case when nul…

Lezas:
Model\User::delete must always return boolean, but here is a case when null is returned.
execute method returns number of affected rows and then trying to reach EOF of this number, null is returned. This causes issues.

### DIFF
--- a/source/Application/Model/User.php
+++ b/source/Application/Model/User.php
@@ -613,9 +613,7 @@ class User extends \oxBase
             $this->deleteAdditionally($sOXIDQuoted);
 
             // and leaving all order related information
-            $rs = $oDb->execute("delete from oxremark where oxparentid = {$sOXIDQuoted} and oxtype !='o'");
-
-            $blDeleted = $rs->EOF;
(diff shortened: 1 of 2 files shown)


keywan-ghadami-oxid:
Hi @Lezas @OXID-eSales/oxid-product-development-team  i did a review, 
The commit does also fix the logic, because the result of deleting an user is true even if their is no order related information to delete. So from my point of view this is also a valid bugfix for 5.x.

@Lezas i suggest to add an assert to the testcase in the usertest:
https://github.com/OXID-eSales/oxideshop_ce/blob/b-6.0-beta.2/tests/Unit/Application/Model/UserTest.php#L1253
to check the return value to be true, that will increase the chance that this pullrequest can be merged soon.

Lezas:
Github doesn't recognize email. 
Will create new pull request with main email, while this one will be closed.

keywan-ghadami-oxid:
Sometimes it help to rebase, sqush and force push things

PR #539, opened 2017-02-12
Use argument unpacking instead of Utilsobject::_getObject

hkreuter:
As OXID eShop requires PHP 5.6 and higher, it is possible to use arguments unpacking.


hkreuter:
Tests were run successfully in CI.

PR #540, opened 2017-02-13
BaseModel _update method fix

Lezas:
Bug ID:6586
Currently this method returns true if number of affected statements was greater than 0 and false on 0. 
This is not good behavior and method should return true on successful query and false on failure.

It's no matter if no rows were affected. It means that data is already updated and true should be returned.

### DIFF
--- a/source/Core/Model/BaseModel.php
+++ b/source/Core/Model/BaseModel.php
@@ -1405,7 +1405,7 @@ class BaseModel extends \oxSuperCfg
 
         $result = $this->executeDatabaseQuery($updateQuery);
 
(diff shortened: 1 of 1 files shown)


Sieg:
Hello @Lezas, thank you for your pull request! Looks like issue is fixed on 2017-03-28, now its eather true on success or DatabaseException on sql error: https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/Model/BaseModel.php#L1379

Feel free to reopen the pull request or create new one if you think current implementation is bad.

PR #544, opened 2017-02-16
fix wrong return type in docblock

alfredbez:
if `$sId` is set, `getCosts()` will return an `oxPrice` object, otherwise it will be an array with all costs


ambrosiustopor:
Thanks!

PR #551, opened 2017-03-14
less code and only one return statement

flow-control:


Sieg:
Hello, thanks for the improvement.

PR #554, opened 2017-03-16
Use ISO-8859-15 encoding for the setup

bertrandjackermann:
PHP 5.6 uses a UTF-8 default header, which causes certain (German) characters to appear broken. For 4.10/5.3.

### DIFF
--- a/source/setup/index.php
+++ b/source/setup/index.php
@@ -20,7 +20,7 @@
  * @version   OXID eShop CE
  */
 
-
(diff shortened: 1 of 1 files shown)


Sieg:
Hello, as this question is not viable for 6.x eshop, and 5.3 have End of life status, i am closing the pull request. Sorry about it, we will try our best to take care of new pull requests faster.

PR #556, opened 2017-03-28
change test-url from www.example.com to localhost

alfredbez:
closes [#6606](https://bugs.oxid-esales.com/view.php?id=6606)

### DIFF
--- a/source/Core/SystemRequirements.php
+++ b/source/Core/SystemRequirements.php
@@ -559,7 +559,7 @@ class SystemRequirements
         $result = 1;
         $iErrNo = 0;
         $sErrStr = '';
-        if ($oRes = @fsockopen('www.example.com', 80, $iErrNo, $sErrStr, 10)) {
+        if ($oRes = @fsockopen('localhost', 80, $iErrNo, $sErrStr, 10)) {
             $result = 2;
             fclose($oRes);
         }


ooxi:
I really don't think this is the correct solution since it assumes a webserver is listening on port 80 on localhost. A common setup contains one host with multiple IP addresses and the webserver listening one those addresses, not `0.0.0.0` (which would include `localhost`).

Wouldn't it be better trying to open the public shop URL? This resolves privacy issues and is more likely to work than `localhost`.

keywan-ghadami-oxid:
I think the shop should check settings like: allow_url_fopen
```
if( ini_get('allow_url_fopen') ) {

}
```
and all known urls from oxid eSales (like the update notification service url).
That way the user can directly see if a php setting or a firewall block something and what exactly.

alfredbez:
> I think the shop should check settings like: allow_url_fopen
>
> ```php
> if( ini_get('allow_url_fopen') ) {
> 
> }
> ```

take a look at [this](https://github.com/OXID-eSales/oxideshop_ce/blob/68d569564cede01b839da8e56ae9390e349214be/source/Core/SystemRequirements.php#L543)

---

We can write something like this:
```php
protected $failedUrls = [];

// ...

public function checkFsockopen()
{
    $urlsToCheck = [
        'http://admin.oxid-esales.com/CE/onlinecheck.php?getlatestversion',
        'http://oxchkversion.oxid-esales.com/webService.php',
        'https://olc.oxid-esales.com/check.php', // needs additional data
        'https://omvn.oxid-esales.com/check.php', // needs additional data
    ];
    $returnCode = 2;
    foreach ($urlsToCheck as $url) {
        if ($res = @fsockopen($url, 80, 0, '', 10)) {
            fclose($res);
        }
        else {
            $returnCode = 1;
            $this->failedUrls[] = $url;
        }
    }

    return $returnCode;
}
```

We can display then all `failedUrls`.

Sieg:
Hey @alfredbez, thanks a lof for the improvement. I have used our OLC host for checking, as that one should be available in most cases. Merged to b-6.1.x, so will be available in the next 6.1 patch.

PR #558, opened 2017-03-29
Catched wild CSS code from the template and put it into an own CSS file.

michaelkeiluweit:
It's better to have the style information in its own place and it has the neat effect, that the CSS code from the admin login page can finally be cached by the browser.

### DIFF
--- a/source/out/admin/src/login.css
+++ b/source/out/admin/src/login.css
@@ -1,0 +1,94 @@
+html, body {
+    background:#fff;
+    height: 100%;
+    padding: 0;
+    margin: 0;
+    font: 11px Trebuchet MS, Tahoma, Verdana, Arial, Helvetica, sans-serif;
+}
+
+div.admin-login-box {
+    border: 1px #FFFFFF solid;
+    border-radius: 5px;
+    width: 332px;
+    padding: 14px 14px 14px;
+    position: absolute;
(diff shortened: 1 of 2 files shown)


source/out/admin/src/login.css
```diff
@@ -0,0 +1,94 @@
...
+    padding: 0;
+    margin: 0;
+    font: 11px Trebuchet MS, Tahoma, Verdana, Arial, Helvetica, sans-serif;
+}
+
+div.admin-login-box {
+    border: 1px #FFFFFF solid;
+    border-radius: 5px;
+    width: 332px;
+    padding: 14px 14px 14px;
```
alfredbez: Maybe we can change this to `padding: 14px;`

source/out/admin/src/login.css
```diff
@@ -0,0 +1,94 @@
...
+    margin: 0 0 5px 0;
+}
+
+.notify {
+    position: fixed;
+    width: 100%;
+    font-size: 16px;
+    color: #fff;
+    background-color: #f77704;
+    padding: 8px 0 8px 0;
```
alfredbez: `padding: 8px 0;`

Sieg:
@michaelkeiluweit, thanks for your pull request!

PR #562, opened 2017-04-26
fix css style in main_facelift for .edittext select.editinput

gerlweb:
small css fix for admin backend


Sieg:
Hi @gerlweb, thank you for your improvement! Looks like this is not viable anymore, as mentioned change already implemented with "Integrate new Admin UI from digidesk backend UI Kit".

PR #564, opened 2017-04-28
changed Hasher from abstract class to interface

alfredbez:
phpcs reported that Hasher@hash declares a return type in the DocBlock without returning anything. The `hash` method is actually an `abstract` method and contains no code. I think it is cleaner to use an interface here.
I'm not sure how to deal with that `BackwardsCompatibilityClassMap` array, see this line:
```
'OxidEsales\\Eshop\\Core\\Hasher' => 'oxhasher',
```


stasiukaitis-saulius:
Hi @alfredbez,

First of all thank you for the idea, but there are few things you could improve.

I. Current solution introduces two possible BC breaks:
  1. A module might extend abstract Hasher and now it is gone.
  2. A module might check with instanceOff if a class created for Sha512Hasher is instance of Hasher 

For this you could:
- Leave old abstract Hasher and deprecate it.
- Introduce a new interface (as you have done).
- Extend abstract and implement interface with Sha512Hasher.

II. Virtual namespace must be used everywhere with a few exceptions. This will make code workable with all three editions.
Replace from:
use OxidEsales\EshopCommunity\Core\Contract\HasherInterface; 
To:
use OxidEsales\Eshop\Core\Contract\HasherInterface;

PR #569, opened 2017-05-05
Update Article.php

arvydasvapsva:
Prepare or at least collect dependencies

### DIFF
--- a/source/Application/Model/Article.php
+++ b/source/Application/Model/Article.php
@@ -488,6 +488,10 @@ class Article extends \OxidEsales\Eshop\Core\Model\MultiLanguageModel implements
                 $this->$sParam = $mValue;
             }
         }
+        
+        // move this and the following dependencies to the constructor and do not use getters inside of methods
+        $this->config = $this->getConfig();
+        
         parent::__construct();
         $this->init('oxarticles');
     }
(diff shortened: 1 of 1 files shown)


source/Application/Model/Article.php
```diff
@@ -488,6 +488,10 @@ public function __construct($aParams = null)
                 $this->$sParam = $mValue;
             }
         }
+        
+        // move this and the following dependencies to the constructor and do not use getters inside of methods
+        $this->config = $this->getConfig();
```
arvydasvapsva: also the following are perfect candidates:

\OxidEsales\Eshop\Core\Registry::getLang()
\OxidEsales\Eshop\Core\Registry::getUtils()
\OxidEsales\Eshop\Core\DatabaseProvider::getDb()

arvydasvapsva: ofc much nicer would be to do this in Base and Multilang classes, so children just access them via ``$this->...``

i guess you got the point. and not BC breaks with this change!

arvydasvapsva: On top - introduce interfaces for all kinds of dependencies:
```php
interface ConfigurableInterface
{
    public function setConfig(ConfigInterface $config);
}

interface DataDependentInterface
{
    public function setDataProvider(DatabaseProviderInterface $dataProvider);
}
...
```

So, depending on Interfaces implemented, ``oxNew()`` can nicely inject dependencies via setters

PR #570, opened 2017-05-05
Update Base.php

arvydasvapsva:
Prepare for the lightweight event dispatcher integration

### DIFF
--- a/source/Core/Base.php
+++ b/source/Core/Base.php
@@ -100,7 +100,11 @@ class Base
      * @return null
      */
     public function __construct()
-    {
+    {                
+        // - introduce at least some dummy event dispatcher, which later cane be injected as a dependency
+        // - follow the idea of http://symfony.com/doc/current/components/event_dispatcher.html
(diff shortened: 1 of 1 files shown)


source/Core/Base.php
```diff
@@ -100,7 +100,11 @@ public function __call($method, $arguments)
      * @return null
      */
     public function __construct()
-    {
+    {                
+        // - introduce at least some dummy event dispatcher, which later cane be injected as a dependency
+        // - follow the idea of http://symfony.com/doc/current/components/event_dispatcher.html
+        // - implement interface based on the symfony dispatcher 
+        $this->eventDispatcher = new SomeEventDispatcher();
```
arvydasvapsva: this way, you can:
- dispatch model specific events, e.g. ArticleUpdated, UserCreated, OrderWasMade etc.
- beautifully decouple cache invalidation code from Article;
- nicely decouple order, user creation email sending from Order, Article etc...

PR #571, opened 2017-05-05
pass along shopid to call to _loadFromDb()

flow-control:
The Parameter $iShopId didn't get passed along to the call to _loadFromDb() which results in null being returned if you try to retrieve a static url from another subshop like so:
oxRegistry::get("oxSeoEncoder")->getStaticUrl("index.php?cl=account_order", $iLang, $iShopId);
If $iShopId is not the same as the active shop, you get null returned.


flow-control:
@hkreuter any news on this?

Sieg:
Hi @flow-control, Thanks a lot for your pull request! Have just merged it.

PR #574, opened 2017-05-08
Update Session.php

arvydasvapsva:
Introduce session interface!


arvydasvapsva:
Thus you can easily implement different session storage/handling strategies. E.g. removing ``sid`` from URLs etc

PR #575, opened 2017-05-08
Update Email.php

arvydasvapsva:
Prepare to decouple from PHPMailer

### DIFF
--- a/source/Core/Email.php
+++ b/source/Core/Email.php
@@ -35,7 +35,7 @@ use Exception;
  * Collects mailing configuration, other parameters, performs mailing functions
  * (newsletters, ordering, registration emails, etc.).
  */
-class Email extends \PHPMailer
+class Email
 {
     /**
      * Default Smtp server port
@@ -288,13 +288,18 @@ class Email extends \PHPMailer
     protected $_oConfig = null;
 
     /**
+     * @var \PHPMailer
+     */
+    private $mailer;
+    
+    /**
      * Class constructor.
      */
     public function __construct()
     {
-        //enabling exception handling in phpMailer class
-        parent::__construct(true);
-
+        // prepare for mailer injection
+        $this->mailer = \PHPMailer(); // woule be even better to implement a wrapper class on top!
+        
         $myConfig = $this->getConfig();
 
         $this->_setMailerPluginDir();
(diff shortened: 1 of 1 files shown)


Sieg:
Hi @arvydasvapsva, Please check the error on travis.

arvydasvapsva:
Hi @Sieg, this PR is a suggestion, to improve. If my idea is clear enough and acceptable, i could spend some time to make it completely working :-) so, what d you think?

keywan-ghadami-oxid:
Hi Arvidas,
maybe others do not know so I like to point that that the following is not official answer from the product team only my personal point of view as project developer:

I think it's a very good idea to use a delegate instead of extending phpmailer, but I am not sure if it is that easy.

Even you use the magic methods to forward method calls to the phpmailer, there might be methods in Email.php that relay on calls from Phpmailer, but they will not be executed anymore because Phpmailer will execute them on its self. 

Maybe you are able to split things into two classes but it will be hard to keep BC while doing so.
So I guess the product team will not be able to give you answer unless they would put a lot of efforts into checking that or you put a lot of effort in proving that. 

So I would suggest to close this pull request and take it on list of ideas for later.
Anyway looking forward to see more cool ideas from you, thank you for sharing it.

arvydasvapsva:
@keywan-ghadami-oxid ``__call()``, ``__set()`` and ``__get()`` should definitely do the job. I could event contribute on my spare time, not to put ``on the list of ideas for later`` :-D

keywan-ghadami-oxid:
@arvydasvapsva I hope I am wrong and it is a easy task so you do not have to spend all your spare time on that. But anyway if it starts to become a bigger task and you need help, you can use https://github.com/TeamPull/oxideshop_ce as central repository while working on it together with other developers (including me).

arvydasvapsva:
@keywan-ghadami-oxid but ``TeamPull`` looks a bit dead thing?

keywan-ghadami-oxid:
@arvydasvapsva it is just nothing to do there, at least nothing i am aware of.
And somtimes git does not tell the true date, commit from yesterday:
https://github.com/OXID-eSales/oxideshop_ce/pull/580/commits/9d5b082f41729f47388fd7d86845263905d09b18

arvydasvapsva:
@keywan-ghadami-oxid does it mean u give me a 👍  so it makes sense for me to continue on the matter?

keywan-ghadami-oxid:
@arvydasvapsva no, but don't get me wrong. I like the idea, but even if it could be done without bc break there is no guarantee that it will be merged, because it is not a feature nor a bugfix, so high risc that it will not be merged in time. But thats again only my personal view, lets wait for more responses.

arvydasvapsva:
@keywan-ghadami-oxid what is the definition of the feature? the idea behind is to decouple the mail client so, that another client (e.g. swiftmailer) could be easily integrated. Furthermore, decoupling would allow to implement async mail sending using third party services like mailgun, mandrill etc

PR #579, opened 2017-05-10
Colon Suffixes moved from language-files to template files

re4jh:
(imho you do not really need to translate them)


kermie:
In general: +1

In the past, stuff like colons where moved to the lang files as you can see here: https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Application/translations/en/lang.php#L37, this lang key for colons should be used in all tpl files in order to spare translations.

re4jh:
COLON is not avaliable in the admins lang.php-Files.
That's why i had to it without the suffix-attribute in the multilang-smarty-tags :-(
In another pull-request we could provide the COLON for the admin-area-lang-files, too.

PR #582, opened 2017-05-23
+- Module (en) is singular form -> German translation should be singular, too!

re4jh:


re4jh:
Dear Travis CI ;)
I just changed an String that is used for text-output.
How could that one fail your tests?

Crease29:
Travis doesn't like English :D

re4jh:
My conspiracy theory is another one:
>> If you change an entry in a german languagefile you have to change the same thing in the englisch pendant at the same time.
.... but in this case this rule makes no sense!

PR #583, opened 2017-05-23
change field length of oxtplblocks.oxmodule to 100 (like oxconfig)

Alpha-Sys:
Change field length of oxtplblocks.oxmodule to 100 (like field oxmodule in oxconfig table). See also bug entry: https://bugs.oxid-esales.com/view.php?id=6090


Sieg:
Hi @Alpha-Sys, Thank you for your pull request! We have added this migration recently: https://github.com/OXID-eSales/oxideshop_ce/commit/b0f9b058ab9e3e94d9fbafdb0d9e24980898a968, does it fix the problem?

Alpha-Sys:
Hi @Sieg, yes, the migration will fix the problem.

PR #584, opened 2017-05-29
CategoryTree: Add oxcmp_basket to the array _aComponentNames as it is used.

michaelkeiluweit:
On a few systems with activated Varnish cache the "body" container of the shop wasn't loaded. You could only see the header and the footer of the shop. After some investigation it became apparent that the object oxcmp_basket had the value "null". As a consequence, a broken HTML oppresed the main body of the shop. That was because a smarty condition couldn't be executed as the object was "null": 

<svg class="shopping-bag-mini[{if $oxcmp_basket->getItemsCount()}] filled[{/if}]" viewBox="0 0 64 64">
ee534/application/views/flow/tpl/widget/header/categorylist.tpl

After adding the component to the "load this components at first" array, it could be loaded on all systems just fine.

### DIFF
--- a/source/Application/Component/Widget/CategoryTree.php
+++ b/source/Application/Component/Widget/CategoryTree.php
@@ -35,11 +35,11 @@ class CategoryTree extends \OxidEsales\Eshop\Application\Component\Widget\Widget
     /**
      * Names of components (classes) that are initiated and executed
      * before any other regular operation.
-     * Cartegory component used in template.
+     * Category and MiniBasket components are used in this template.
      *
      * @var array
      */
-    protected $_aComponentNames = array('oxcmp_categories' => 1);
+    protected $_aComponentNames = array('oxcmp_categories' => 1, 'oxcmp_basket' => 1);
 
     /**
      * Current class template name.


Sieg:
Hi @michaelkeiluweit, can't we just improve the condition which is failing? To check if the component is not null?

michaelkeiluweit:
Hi @Sieg . Maybe. But in my opinion the object should be there, as the widget assumes exactly that. The widget needs to know, how many products are in the basket and the minibasket has this information.

Sieg:
Yes @michaelkeiluweit, you are right.

stasiukaitis-saulius:
Hi guys

This approach would not work as Varnish would cache menu widget with result from basket. This means that buyer would see amount in basket independent of what he actually has in his basket but dependent from what was cached in Varnish.

We are working with this bug atm. story: ESDEV-4774.

Currently changes was done in Flow theme and Enterprise repository. You can find branches by the story id. Our goal is to merge bug fix back to master before next release.

Cheers

michaelkeiluweit:
Alright, then I'll close this PR.

PR #586, opened 2017-07-03
Integrate new Admin UI from digidesk backend UI Kit

pnkz:


robertblank:
Merged in this commit: https://github.com/OXID-eSales/oxideshop_ce/commit/662d3beaa4f1f5f935189ca804f71b5b8164782a

robertblank:
Hi Pascal, 
just for the record.
As it should usually be avoided, but sometimes there are good reasons to do so:
Can you shortly explain the concept behind using !important in those CSS files?

pnkz:
Hey Robert,
i wanted to let the legacy css files untouched, that someone can disable the facelift styles and have the old UI back. So at some points i had to overwrite the rules with "!important" in the facelift css files.

PR #592, opened 2017-08-04
only getAdminTplLanguageArray() when isAdmin()

mattvb91:
getObjectTplLanguage() will return wrong language ID due to missing isAdmin() check. If the language doesn't exist on admin it can cause issues.


Sieg:
Hi @mattvb91, Thanks for your pull request! Could you add a test which covers the "admin" case of the method?

mattvb91:
Hi @Sieg sure thing i'll add it over the coming days. I actually didn't check your bug tracker if this is a reported issue or not so I cant tell you. I didn't open an issue as it was a pretty easy find.

Sieg:
Hello @mattvb91, any news here? :)

Sieg:
Thank you @mattvb91, merged on 98532f8.

PR #594, opened 2017-08-10
Update oxseoencoder.php

BattleMoench:
i think, there was a minor bug in oxseoencoder, which should be fixed with this.. SEO-URLs are stored separately for every subshop. In oxseoencoder->_loadFromDb() SEO-Urls, which are marked as expired, were copied to seohistory-Table for given ObjectID and shopID, but the expired-Flag was unset for given ObjectID and every Subshop (given shopId wasnt used). This should work better now.


Sieg:
Hi @BattleMoench, thanks for your pull request! The changes are looking good, but could you please:
* Rebase it on our master or b-6.x branch.
* Do you have the exact bug number by chance? While scanning, I coulnd't find it.

Sieg:
Thank you, have added the line as we were discussing privately.

PR #597, opened 2017-09-08
removed undefined $oCur

vanilla-thunder:
$oLang->formatCurrency() can get active currency if there is no currency passed.

also, there are some translation idents missing
image
(tested in demoshop)


Sieg:
Thank you @vanilla-thunder for your PR. We will take only critical bug fixes to v5.3. 

Marat confirmed that Problem is not viable on v6 anymore, so i am just closing the PR and no further actions needed regarding the topic.

PR #603, opened 2017-10-09
fix for b6703

michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=6703
Fixed the "robots.txt is case sensitive" bug and updated the path of the english terms and conditions site.


iegupov:
Thank you for your pull request. It was recently merged to the master and cherry picked to 6.x on 54ca55275f8f9f11396bff628e39c5dc8be578d2

PR #608, opened 2017-11-07
pass correct oxseo type for manufacturerUrl

mattvb91:
On line 69 _saveToDb is called with the `oxmanufacturer` type:

https://github.com/OXID-eSales/oxideshop_ce/blob/f79cadfd4269830b5eebbe82b6d1ddd1808a37af/source/Application/Model/SeoEncoderManufacturer.php#L69

However in the `getManufacturerPageUrl()` method the type is passed as `oxmanufacturers` which results in never being able to read the result from DB & always writing on every request instead even when a result already exists.

### DIFF
--- a/source/Application/Model/SeoEncoderManufacturer.php
+++ b/source/Application/Model/SeoEncoderManufacturer.php
@@ -94,10 +94,10 @@ class SeoEncoderManufacturer extends \OxidEsales\Eshop\Core\SeoEncoder
         $seoUrl = $this->getManufacturerUri($manufacturer, $languageId);
 
         if ($isFixed === null) {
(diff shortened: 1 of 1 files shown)


proudcommerce:
(Y)

hkreuter:
Hi mattvb91 , 
thank you very much for your contribution. Unfortunately, we cannot take your pull request. It might look like a typo, but there is indeed a difference between 'oxmanufacturer' and 'oxmanufacturers' oxseo.oxtype. We already are aware of this issue, see https://bugs.oxid-esales.com/view.php?id=6694 and will fix it in a future release.

PR #612, opened 2017-12-06
added commit/rollback for unfinished transactions in order

KristianH:
take a look at https://bugs.oxid-esales.com/view.php?id=6736

this commit could be a temporary solution until the finalizeOrder method is refactored.


stasiukaitis-saulius:
Hi @KristianH,

Thank you for this pull request. We are working on the same issue at the moment but with a slightly different solution. After discussions with domain experts it was decided for now to remove the transaction as information might be needed even when Payment fails. Especially if information about Payment is partly saved in 3rd party systems.

https://github.com/OXID-eSales/oxideshop_ce/commit/e2a7bba9dc035b8b8537345e4d4eac4f792302a6 

Feel free to let us know if you have some more hints about this issue.

KristianH:
I'm looking forward to see your solution :+1:

PR #614, opened 2017-12-06
display full field names in filter dropdowns in admin

vanilla-thunder:
since its almost 2018 and screen resolution and size increased over last 20 years, we could display full field names in filter dropdowns 
image


Sieg:
Thanks, merged.

PR #616, opened 2017-12-08
Added cache for getAttributesDisplayableInBasket()

Zockman:
`getAttributes()` and `getAttributesDisplayableInBasket()` both use `$this->_oAttributeList` as a result cache.  
This leads to wrong results if both methods are called during the same request.

This change adds `$this->_oAttributeListDisplayableInBasket` to separate the caches.

### DIFF
--- a/source/Application/Model/Article.php
+++ b/source/Application/Model/Article.php
@@ -206,6 +206,11 @@ class Article extends \OxidEsales\Eshop\Core\Model\MultiLanguageModel implements
     protected $_oAttributeList = null;
 
     /**
+     * Object holding the list of attributes and attribute values associated with this article and displayable in basket
+     */
+    protected $_oAttributeListDisplayableInBasket = null;
+    
+    /**
      * Indicates whether the price is "From" price
      *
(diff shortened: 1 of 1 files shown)


Sieg:
Hello @Zockman, thanks for pull request, but some little improvements needed: please use new coding style for new variables, something like: $this->basketAttributesList.

tabsl:
> please use new coding style for new variables, something like: $this->basketAttributesList.

When was that decided? Where can we found more information about this adjustment?

Sieg:
Hello, please check the https://oxidforge.org/en/coding-standards.html.

tabsl:
@Sieg haha --> The so called “Hungarian notation” is used for variables and attributes.

Sieg:
Yes, exception for 5.2, what's wrong with it? Sorry, didn't get it:)

Sieg:
Hello @Zockman, i have merged your changes with some fixes from my side. Thanks for your contribution. Merge points: c218744 and 1796d9c1

PR #617, opened 2017-12-12
use shallow clone to reduce download size by 75%

vanilla-thunder:
at the current state:
git clone https://github.com/OXID-eSales/oxideshop_ce.git
downloads 140 megabytes of files and history of their changes.
by using --depth=1 you can make a shallow clone, which contains only current state of the file,
It reduces total download size to 33 megabytes


Sieg:
Hello @vanilla-thunder, Thank you for the pull request. We had this point mentioned in our contributing.md files, but we have improved our readme.md and contributing.md files now, so people could be more up to date how the installation and contributing should be done. Feel free to check the cb0b00d commit.

PR #618, opened 2017-12-26
bugfix #6761

vanilla-thunder:
https://bugs.oxid-esales.com/view.php?id=6761


vanilla-thunder:
also, this fix should be applied to b-5.3-ce

Sieg:
Thank's Marat.  Merged to b-6.x and master.

PR #620, opened 2018-01-08
fixed entity error with &amp; value

nymo:
In Backend category main when you select a sub category which has a & char in title the char is displayed as &amp;. This commit should fix it.


Sieg:
Thank you @nymo for your Pull request. I have merged it recently with some improvements on the top.

PR #621, opened 2018-01-08
some maintenance work for OXIDforge links:

kermie:
* replaced http with https
* replaced /en/installation.html with /en/system-requirements


Sieg:
Merged to 6.x and master. Thanks.

PR #625, opened 2018-01-31
use correct setting for manufacturer slider

Alpha-Sys:


Alpha-Sys:
See bugtracker: https://bugs.oxid-esales.com/view.php?id=6783

PR #630, opened 2018-02-21
set server collations during migration

KristianH:
After update of a shop to actual v6.x.x there can be this message during "php vendor/bin/oe-eshop-db_migrate migrations:migrate":

```
Migration 20171018144650 failed during Execution. Error An exception occurred while executing 'INSERT INTO `oxconfig` (`OXID`, `OXSHOPID`, `OXMODULE`, `OXVARNAME`, `OXVARTYPE`, `OXVARVALUE`)
          SELECT
            CONCAT(SUBSTRING(`OXID`,1, 16),SUBSTRING(REPLACE( UUID( ) ,  '-',  '' ), 17,32)) AS `OXID`,
            `OXSHOPID`,
            `OXMODULE`,
            "blSendTechnicalInformationToOxid" AS `OXVARNAME`,
            `OXVARTYPE`,
            `OXVARVALUE`
          FROM `oxconfig`
          WHERE `OXVARNAME` = 'blLoadDynContents'
          AND NOT EXISTS (
            SELECT `OXVARNAME` FROM `oxconfig` WHERE `OXVARNAME` = 'blSendTechnicalInformationToOxid'
          );':

SQLSTATE[HY000]: General error: 1270 Illegal mix of collations (utf8_general_ci,COERCIBLE), (utf8_unicode_ci,COERCIBLE), (utf8_unicode_ci,COERCIBLE) for operation 'replace'
```

### DIFF
--- a/source/migration/data/Version20170718124421.php
+++ b/source/migration/data/Version20170718124421.php
@@ -20,6 +20,10 @@ class Version20170718124421 extends AbstractMigration
      */
     public function up(Schema $schema)
     {
+        /// set server collation variables
+        $this->addSql("SET @@session.collation_connection = 'utf8_general_ci';");
+        $this->addSql("SET @@session.collation_server = 'utf8_general_ci';");
+
         $this->addSql("ALTER TABLE `oxtplblocks` 
           CHANGE `OXMODULE` `OXMODULE` varchar(100) 
           character set latin1 collate latin1_general_ci NOT NULL 
--- a/source/migration/data/Version20171018144650.php
+++ b/source/migration/data/Version20171018144650.php
@@ -19,6 +19,10 @@ class Version20171018144650 extends AbstractMigration
      */
     public function up(Schema $schema)
     {
+        /// set server collation variables
+        $this->addSql("SET @@session.collation_connection = 'utf8_general_ci';");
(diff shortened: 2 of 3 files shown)


Sieg:
Hello @KristianH, is this really required? What is the configuration of your MySQL server for those variables? Maybe it is better to configure server default values, to be sure other queries run correctly?

KristianH:
What if you can't change the server default values for security reasons?
The alternative is to rebuild the queries with using "COLLATE" to avoid the errors.

Sieg:
Hello @KristianH, i have just checked with team, we already are working on some solution to this problem. The quick workaround for fixing the current issue - try improving the `migrations-db.php` file like this:

```
return [
   ...
    'charset' => utf8
   // not necessarily but the shop does the same:
    'driverOptions' => [
        \PDO::MYSQL_ATTR_INIT_COMMAND => "SET @@SESSION.sql_mode=''"
    ]
];
```

KristianH:
I close this PR, the fix is already provided in:
https://github.com/OXID-eSales/oxideshop-doctrine-migration-wrapper/commit/cc9d230b8098c5ea53e4251e1b7951df565bee3b

PR #631, opened 2018-02-23
Removed unused parameter

michaelkeiluweit:
It is nowhere used.


Sieg:
Thank you @michaelkeiluweit, have removed the parameter as you suggested. Its deprecated in b-6.x and completely removed in master.

PR #632, opened 2018-02-27
admin selectlist: add filter for oxident

Alpha-Sys:
Add a new input field to filter for oxident field.


Sieg:
Thank you @Alpha-Sys, have just merged to b-6.x and master.

PR #633, opened 2018-02-28
add database changes for large module lists in oxconfig table

SeifertDaniel:
"feature" in https://bugs.oxid-esales.com/view.php?id=6164


Sieg:
Hi @SeifertDaniel, Thanks for your pull request. There is no need to update the basic schema, as migrations should take care of it. I have merged your migration at 03b15d99303a68086f2bf119cba2c41ce57102ff and at 04ab73938283add6d8a55d9fd2c0ad3122f8e194 it came to master.

PR #637, opened 2018-03-15
Adjusted the old path to the logo.

michaelkeiluweit:


michaelkeiluweit:
I used the image "logo_oxid.png" as there is no default image "logo.png" anymore. But I think it would be better to have a default logo.png therefore the shop owner can replace it by his own logo. The name logo.png is much more obvious than "logo_oxid.png"

Sieg:
Thank you for fixing this, @michaelkeiluweit. Have merged your PR.

PR #636, opened 2018-03-09
Change function declarations to protected

cpjolly:
The methods isSerializedBasketValid, isClassInSerializedObject, isClassOrNullInSerializedObjectAfterField and isUnserializedBasketValid need to accessible to classes that extend the public Session::getBasket method.

PR #638, opened 2018-03-18
Set canUserAccountBeDeleted() visibility to public

cpjolly:
The method canUserAccountBeDeleted() needs to have either public or protected visibility, so that the deleteAccount() method can be modified by modules that extend AccountController.php

### DIFF
--- a/source/Application/Controller/AccountController.php
+++ b/source/Application/Controller/AccountController.php
@@ -387,7 +387,7 @@ class AccountController extends \OxidEsales\Eshop\Application\Controller\Fronten
      *
      * @return bool
      */
-    private function canUserAccountBeDeleted()
+    public function canUserAccountBeDeleted()
     {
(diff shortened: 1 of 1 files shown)


Sieg:
Hello @cpjolly, There is isUserAllowedToDeleteOwnAccount near by, it is public, why do you want this one to be public as well?

cpjolly:
Hi @Sieg ,

if canUserAccountBeDeleted() is private, it is impossible to override the public method deleteAccount() with a version that still uses canUserAccountBeDeleted().

This means if you only want to very slightly change the behaviour of deleteAccount() you need to write your own version of canUserAccountBeDeleted().

This example illustrates why all methods on classes that can be extended should be either public or protected, not private. This is the exact same issue as #636 

In the past, all extendable OXID classes obeyed this principle.

PR #639, opened 2018-03-23
Doctrine's QueryBuilder feature

tmloberon:
Open [QueryBuilder feature](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/query-builder.html) of Doctrine, please!

Than we can write code like:

```php
$database     = DatabaseProvider::getDb(DatabaseProvider::FETCH_MODE_ASSOC);
$flexibleSql = $database->createQueryBuilder();

$flexibleSql->select('*')
            ->from(getViewName('oxarticles'))
            ->where('oxtitle like :value')
            ->setParameter(':value', "%Sessel%");

$result = $database->getAll($flexibleSql->getSQL(), $flexibleSql->getParameters());
```

We have a flexible sql object. To change it easy in other function like:

```php
...
addActiveSnippet($flexibleSql);
...
$result = $database->getAll($flexibleSql->getSQL(), $flexibleSql->getParameters());

function addActiveSnippet(QueryBuilder $flexibleSql) {
    $flexibleSql->andWhere('oxactive = 1');
}
```

### DIFF
--- a/source/Core/Database/Adapter/DatabaseInterface.php
+++ b/source/Core/Database/Adapter/DatabaseInterface.php
@@ -400,4 +400,11 @@ interface DatabaseInterface
      * @return string|int Row ID
      */
     public function getLastInsertId();
+
+    /**
+     * A QueryBuilder to dynamically create SQL queries.
+     *
+     * @return object
+     */
+    public function createQueryBuilder();
 }
--- a/source/Core/Database/Adapter/Doctrine/Database.php
+++ b/source/Core/Database/Adapter/Doctrine/Database.php
@@ -1418,4 +1418,19 @@ class Database implements DatabaseInterface
 
         return $code;
     }
+
+    /**
+     * A QueryBuilder to dynamically create SQL queries. Use origin Doctrine's QueryBuilder
+     *
+     * @see http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/query-builder.html
+     * @return \Doctrine\DBAL\Query\QueryBuilder
+     */
+    public function createQueryBuilder()
+    {
+        if (is_null($this->connection)) {
+            $this->connect();
+        }
+
+        return $this->getConnection()->createQueryBuilder();
+    }
 }
--- a/tests/Integration/Core/Database/Adapter/Doctrine/DatabaseTest.php
+++ b/tests/Integration/Core/Database/Adapter/Doctrine/DatabaseTest.php
@@ -572,6 +572,12 @@ class DatabaseTest extends DatabaseInterfaceImplementationTest
         );
     }
 
+    public function testGetInstanceOfDoctrineQueryBuilder()
+    {
+        $actualResult = $this->database->createQueryBuilder();
+        $this->assertInstanceOf('\\Doctrine\\DBAL\\Query\\QueryBuilder', $actualResult);
+    }
+
     /**
      * Assert, that the given array is unique.
      *


patrikkernke:
That would be a great feature 👍 
Cleaner and readable code!

smxsm:
That would be awesome, please approve this! 👍

mattvb91:
+1 this is a must have

mtkoltan:
Sorry, but we have to reject this pull request due to backward compatibility issues. As you know we replaced the ADOdb light framework we used previously in the shop with Doctrine; and a lot of you are annoyed that we hid the Doctrine connection and with this also the possibility to use the query builder. We did not do this lightly, but for a reason: The ResultSet returned by Doctrine behaves differently as the ResultSet from ADOdb light. While ADOdb light places the cursor already on the first row of the ResultSet, for Doctrine you have to call fetchRow() once before you get the first result. Of course we could have changed all the places where we use such a ResultSet - but what about the thousands of modules out there in projects? Would all module developers be responsible enough to read our blog posts and update their modules accordingly? Definitely not.

So we decided to hide the Doctrine connection for now to avoid confusion with two different kinds of ResultSets - at least in the code that might have already been extended by some modules. Otherwise we inevitably would have introduced off by one errors that are really hard to detect. If we now opened up the Doctrine QueryBuilder, we would have two different ResultSets floating around in the code and developers would really be confused.

This does not mean that we think it is a bad idea to use Doctrines query builder, on the contrary. We recently introduced a new namespace in the OXID code base named Internal. Within this Internal namespace we will use the Doctrine connection and also the QueryBuilder. And in the future we will provide you as OXID developers new means to extend this Internal code (using events and dependency injection). But for now we will keep the Doctrine connection and QueryBuilder hidden to avoid chaos with two different types of ResultSets.

We regret this decision, but I hope you understand our decision in favour of the stability of your projects and those of others.

tmloberon:
Ich habe es geahnt, dass es nicht geöffnet wird. Das ist schade.

flow-control:
buuuh

flow-control:
I dont see the backward compatibility issue here. The changes **do not change existing behavior**, they add new methods, which have not been there in the past.

mtkoltan:
Yes, for the moment it would just add behaviour. But if you add behaviour, it will be used. And then you will have two different ResultSet objects in the code, that behave differently. And this will lead to trouble we want to avoid. I'm really sorry.

Betazoid:
> you will have two different ResultSet objects in the code, that behave differently
How about documenting different behaviour? Cross-reference and emphasize it. This isn't unusual in projects maintaining compatibility while moving forward. 

> Yes, for the moment it would just add behaviour. But if you add behaviour, it will be used.
What do you mean by this? Isn't this a big part of the definition of _feature_ and _improvement_?

Closing, this a really poor excuse.

mattvb91:
@BenjaminJoerger @Sieg your bugtracker is broken and this is the most relevant pull request I could find... (Could you allow opening issues here instead of using the oxid bugtracker (Broken for months)).

There are 2 SQL Injection possibilities in https://github.com/OXID-eSales/oxideshop_ce/blob/ec33d6d7cab3d49632f315aaaa2ab416055a7799/source/Application/Controller/Admin/CategoryOrderAjax.php#L120 as $soxId is not escaped when in admin.

Sieg:
@mattvb91, what is broken for you in bugtracker? It works fine (https://bugs.oxid-esales.com)

mattvb91:
@Sieg we have had multiple devs attempt to open bug reports and all receive either blank page when submitting or redirect to bug report form with no error message. Feel free to open a bug with the sql injection above ^^

Sieg:
@mattvb91, i have created an issue in bugtrack, https://bugs.oxid-esales.com/view.php?id=6890, temporary marked as private.

Sieg:
@mattvb91, i have just created a new account in bugtrack, reported an issue with no errors or redirects as you say, works just fine. Please tell me what browser or technologies you use for us to reproduce your issue.

PR #642, opened 2018-04-05
Fixed notices and performance in isVatIncluded

MaxBoeh:
I rearranged the method FrontendController::isVatIncluded to:
(If the user is not logged in)
- Avoid loading Country instance with id "" (empty string)
- Avoid two php notices per call (Country was not successfully loaded)
> Trying to get property of non-object in source/Application/Controller/FrontendController.php line 2894

(If the user is logged in)
- Avoid repeated evaluation on subsequent calls, including loading the same Country object

(In either case)
- Avoid creating unnecessary objects if the method result only depends on the configuration

I'm done with my changes and would appreciate your review. However I think I should have done them on base of branch b-6.x...

### DIFF
--- a/source/Application/Controller/FrontendController.php
+++ b/source/Application/Controller/FrontendController.php
@@ -2891,16 +2891,6 @@ class FrontendController extends \OxidEsales\Eshop\Core\Controller\BaseControlle
     public function isVatIncluded()
     {
         $config = $this->getConfig();
-        $user = $this->getUser();
-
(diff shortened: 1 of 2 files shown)


Sieg:
Thank you @MaxBoeh, i have merged your improvements recently at 0ced287119258cdd6ac4b1d1e6256a8457a16a5d. I have squashed some commits into one, and improved commit titles a little, hope its fine.

PR #643, opened 2018-04-13
Streamline 404 use case

flow-control:
This one was the last occurrence of a direct call to `Utils::handlePageNotFoundError()`, which should also use `error_404_handler()`


Sieg:
Hello @flow-control, Thank you for your improvement. Merged at 54e83e79007e090ea09ae90335930d802f5adb58

PR #644, opened 2018-04-18
Set PHP 7.1 as maximal recommended version

Trainmaster:
I think there is no reason for not setting PHP 7.1 as maximal recommended version since CI passes using that PHP version.

### DIFF
--- a/source/Core/SystemRequirements.php
+++ b/source/Core/SystemRequirements.php
@@ -557,7 +557,7 @@ class SystemRequirements
 
         $minimalRequiredVersion = '5.5.0';
         $minimalRecommendedVersion = '5.6.0';
-        $maximalRecommendedVersion = '7.0.9999';
+        $maximalRecommendedVersion = '7.1.9999';
 
         $installedPhpVersion = $this->getPhpVersion();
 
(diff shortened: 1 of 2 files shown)


cpjolly:
Why not 7.2 ?
"Active Support" for 7.1 ends in 12.2018
http://php.net/supported-versions.php

mattvb91:
:+1:

keywan-ghadami-oxid:
FYI : 7.2 was requested by some hosters because it now default in some environments.

der-onkel:
+1 for PHP 7.2, because it's the default Version un Ubuntu 18.04LTS

M4rt1n17:
I would appreciate a full PHP 7.2.x Support very much!
When I change now, from PHP 7.1.x to PHP 7.2.x I get this Warning, when loging in the Backend:

php 7 2 x

Problem is: the Shop notices that "ini_set" is disabled in PHP 7.2.x (as far as I know)

php 7 2 x systemgesundheit

So if PHP 7.2.x is set as recommended PHP version it also should not throw any red warningat System Health.

Also MariaDB is not fully supported (I'm running on "10.0.32-0+deb8u1") and got a orange warning

Sieg:
Hello guys, for sure we will support higher php versions! It's not that simple for us to just change one number and that's it. We need to prepare the infrastructure to run our builds with those higher versions and ensure everything works fine. It takes time. I will let you know if there will be progress on this one, or better - When there will be progress on it :)

Sieg:
Hello guys, so for OXID 6.1 we have:

        $minimalRequiredVersion = '7.0.0';
        $minimalRecommendedVersion = '7.0.0';
        $maximalRecommendedVersion = '7.1.9999';

We are not yet ready for 7.2, but team is working on fixing those.

I think, the question is closed for now. Thank you everybody for your interest in the question!

PR #645, opened 2018-04-22
ExceptionHandler compatibility with php7

MaxBoeh:
ExceptionHandler can now deal with php 7 Errors (warnings, notices). Before the change they lead to catchable fatal errors because of the typehint for \Exception.

I also have a related question: I'm one of those who enable the full error_reporting especially in production as warnings / notices are good indicators of bugs. I would really appreciate it if we could get rid of php warnings and notices, would you be interested in related pull requests?


Sieg:
Hello @MaxBoeh, regarding your question of warnings/notices - Yes, for sure we would be happy to get such improvements! Small pull requests with easy to reproduce instructions/descriptions are merged quite fast.

I like this pull request, but will discuss with a team first!

PR #647, opened 2018-05-15
Update Language.php

mattvb91:
the array_merge causes array keys to be lost. Which as a result breaks view generation.

Example before array_merge:

```php
$aLanguages = [
    0 => 'de',
    9 => 'es',
]
```

after: 

```php
$aLanguages = [
    0 => 'de',
    1 => 'es',
]
```

This causes views for example in oxv_oxarticles_es to be generated with OXTITLE_1 instead of OXTITLE_9.

Does anyone have a cleaner solution or was there a change made recently to how view generation is handled?

### DIFF
--- a/source/Core/Language.php
+++ b/source/Core/Language.php
@@ -1326,7 +1326,7 @@ class Language extends \OxidEsales\Eshop\Core\Base
                 $aLanguages = $this->_getLanguageIdsFromLanguagesArray($aConfigLanguages);
             }
 
-            $aConfigDecodedValues = array_unique(array_merge($aConfigDecodedValues, $aLanguages));
+            $aConfigDecodedValues = array_unique($aConfigDecodedValues + $aLanguages);
         }
 
         return $aConfigDecodedValues;


keywan-ghadami-oxid:
@reviewers:
I found a documented sample of that behavior of array_merge:
https://www.w3schools.com/php/showphp.asp?filename=demo_func_array_merge3
there is also a accepted answer on stack overflow for this topic is:
https://stackoverflow.com/questions/17462354/how-to-merge-array-and-preserve-keys

I hope that additional information may help to reduce effort for the review.

Sieg:
Hi @mattvb91, what you say regarding the difference in keys by using the array_merge is correct, but, we were not able to reproduce the breaking of the views in the shop. Could you please give us the example or even better - the test, to reproduce the problem? Do you have any bugs reported to our bugtrack? Could you please link it if there is one?

Sieg:
Hi @mattvb91, i have reported a bug, by what i was able to find yesterday, please add more information there, if you have some other cases too: https://bugs.oxid-esales.com/view.php?id=6846

The pull request fixes this problem, but will introduce another problem - the method is used to get a list of languages in most cases, and the result will be wrong with array(0=>'lt') + array(0=>'de', 1=>'en'), because keys matters for this operation.

mattvb91:
Hey @Sieg sorry I dont have enough time to dig deeper into this at the moment but I will update if I can. Like you suggested on the issue it may just be that the function name is wrong and should be different.

Sieg:
Hi @mattvb91, i should create the PR as it cannot be taken directly because of already described problems. It will not be easy to fix the problem, but feel free to open new one or reopen current one if you have some improvements which handles the problem in more global view. The bug is open and pull request (even closed, with current solution) is linked to the bug.

PR #649, opened 2018-05-19
Added php warnings for not executed db statements

MaxBoeh:
Reading DB statements are not executed when used with the wrong doctrine methods. While migrating from Oxid 4 and sometimes during development we experience the silent return of empty results, therefore I added a warning whenever a statement is not executed.
I preferred the warning over a simple log entry as it can be handled and automatically contains a stack trance.

(We also had additional trouble as we are heavily using WITH clauses)

### DIFF
--- a/source/Application/Model/Order.php
+++ b/source/Application/Model/Order.php
@@ -899,7 +899,7 @@ class Order extends \OxidEsales\Eshop\Core\Model\BaseModel
      * deletes order and returns * error code 2.
      *
      * @param \OxidEsales\Eshop\Application\Model\Basket $oBasket      basket object
-     * @param object                                              $oUserpayment user payment object
+     * @param object                                     $oUserpayment user payment object
      *
      * @return  integer 2 or an error code
      */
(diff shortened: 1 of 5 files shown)


Sieg:
Hi @MaxBoeh, thanks for the Pull request! You are right, sounds good to throw some warnings on method missuse, but we would like some little improvements here: We have introduced the psr logger recently, so lets use it for logging those warnings, but not throw them directly (Registry::getLogger();..->warning();). Could you check it and improve?

MaxBoeh:
Hi @Sieg, thanks for your feedback, I changed it to logging.
I would like to have some additional information in the log entry to know where the method was called so one can directly fix it without the need to start debugging. How can this be achieved?

Sieg:
Hi @MaxBoeh, regarding your last question - Maybe initializing and logging the exception, but not simple string?

Sieg:
I will add [debug_backtrace()] as a context for the warning method, so we will see the trace.

Sieg:
Hi @MaxBoeh, thanks for your pull request. Merged to b-6.x recently on ad6a0645cebbb73a09597e6ace29ab6865216bb8

PR #650, opened 2018-05-19
Fixed several notices caused by uninitialized / null variables

MaxBoeh:
Fixed most of the php notices I've found so far.

However I've found one issue I'm not sure how to deal with: In _FrontendController::isVatIncluded_ the database field _oxuser__oxustidstatus_ is used - but it doesn't seem to exist in the community edition. I've found an eight years old and resolved [ticket](https://bugs.oxid-esales.com/view.php?id=1653), but it is not mentioned how it was resolved / if that's intentional.

Can the database field be added to CE or the field be removed from production code?


MaxBoeh:
Obviously I messed it up merging with the master. Sorry guys -.-
Any pointers to avoid that are very appreciated.

Sieg:
Hi @MaxBoeh, i saw you are preparing those commits as new PR, right?

PR #651, opened 2018-06-03
Fix variable name

Indianer3c:
This pull request fixes the confusing issue "not stopped correctly!" if everything went right.

### DIFF
--- a/source/Core/DebugInfo.php
+++ b/source/Core/DebugInfo.php
@@ -79,7 +79,7 @@ class DebugInfo
     {
         $log = 'Execution time:' . round($dTotalTime, 4) . '<br />';
         global $aProfileTimes;
-        global $executionCounts;
+        global $aExecutionCounts;
         if (is_array($aProfileTimes)) {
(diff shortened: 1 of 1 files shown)


Sieg:
Hi @Indianer3c, could you please write more about the problem? By our new coding style, we do not use "Hungarian notation" for new code/changes anymore, so your suggested change does not fit.

Indianer3c:
Hi @Sieg :) ah okay, but in source/overridablefunctions.php in function startProfile() is the variable $aExecutionCounts used. E.g. when I set $this->iDebug = -1; into config.inc.php then I always get the message "not stopped correctly!". When "Hungarian notation" not use anymore then you can change the variable $aExecutionCounts into startProfile() to $executionCounts please?

Sieg:
Hi @Indianer3c, thanks for clarifying the problem! Yah, i will update the variable name in startProfile, let's move forward, but not backward 🗡

Sieg:
Hi @Indianer3c, thanks for your pull request. Fixed in b-6.x recently (15f716140a02816642aa7bb4532ef16af96ad876)

PR #652, opened 2018-06-04
Fixed several notices caused by uninitialized / null variables (second try)

MaxBoeh:
(After messing up pull request #650, here comes the second try. I apologize for any inconvenience)

Fixed most of the php notices I've found so far.

However I've found one issue I'm not sure how to deal with: In FrontendController::isVatIncluded the database field oxuser__oxustidstatus is used - but it doesn't seem to exist in the community edition. I've found an eight years old and resolved ticket, but it is not mentioned how it was resolved / if that's intentional.

Can the database field be added to CE or the field be removed from production code?


Sieg:
Thanks @MaxBoeh, merged at b96c3d565e8afc867dccd4fa61f1b9e9aff74857.

Sieg:
Regarding the "oxuser__oxustidstatus" field - cannot comment anything on it. Please report an issue in bugtrack, if something is missing and not working.

PR #653, opened 2018-06-06
Removed oxusername (email) from getTitle() method.

kavatari:
This is very qustionable under the consideration of the GDPR law in europe - especially when your shop is tracked by google analytics, the whole email address will be shown in google reports.


Sieg:
Hello @kavatari, 

Thank you very much for your interest in the question! You are right, there are better approaches than putting an email address into the title. Just removing the email is not a valid solution as the current user needs to see with which account he is logged in.

Due to this, we can not accept your PR. Btw, most shop owners usually adjust this template to their needs: showing the last name only, or first name, last name and company name - so many options.

PR #655, opened 2018-06-23
remove smtp connect check

flow-control:


keywan-ghadami-oxid:
You removing the smtp connection check, why? - is that needed because in case of error the fallback is wrong or because the check does not work at all (my because firewall or Mailserver does not accept this test) ?

flow-control:
Hey there, this merge request is not complete yet. But to clarify: I would not check for the host to be correct in this state, but upon entering the URL in the backend

PR #656, opened 2018-06-25
avoid race conditions on a slow db with a big basket

keywan-ghadami-oxid:
avoid race conditions
using a transaction to make the action atomic, (saving the changed basket or saving it not)
 (It was reported that randomly some basket items got lost on big baskets > 350 items ((and slow database))
To avoid long running locks, the new implementation writes the changed basket with a new id and name and after writing all basket items to the databases it updates the name and the current id within a transaction.

TODO:
- [ ] contact support

### DIFF
--- a/source/Application/Model/Basket.php
+++ b/source/Application/Model/Basket.php
@@ -1691,10 +1691,11 @@ class Basket extends \OxidEsales\Eshop\Core\Base
                 $oSavedBasket->oxuserbaskets__oxtitle = new \oxField('savedbasket');
 
                 //switching baskets within transaction to avoid race conditions
-                oxDb::getDb()->startTransaction();
+                $database = \OxidEsales\Eshop\Core\DatabaseProvider::getDb();
+                $database->startTransaction();
                 $oldSavedBasket->save();
                 $oSavedBasket->save();
-                oxDb::getDb()->commitTransaction();
+                $databse->commitTransaction();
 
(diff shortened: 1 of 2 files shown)


Sieg:
Hello, is there a bug reported with this problem? Please link it, if not, please, give more information on how to reproduce the issue. 

Do we really need this second basket? Cannot it just be done with opening the transaction at the start, and committing all the old logic at the end?

keywan-ghadami-oxid:
@Sieg that was also my first suggestion, but having a huge transaction comes with the risc blocking more resources and by that a higher chance of creating deadlocks or making things go slowing. 

As I was only doing remote support on a productive system with real users, I do not know how to provide more details, I am going to contact support team to check if it is a known issue.

Sieg:
Hi @keywan-ghadami-oxid, thanks for your explanations. Can you be more precise what do you mean by "more resources"? I missed something probably, but i see only ones that active user should have access to, no? so how it can be broken here..?

Could you please check the Travis too?, something does not fit together.

Sieg:
Hi @keywan-ghadami-oxid, just curious why you have closed the pull request, not enough time to check the tests of you found something bad?. I have just realized, that actually you dont need to update that old basket at all, you are going to remove it anyway, so the code can be even shorter and more clear.

keywan-ghadami-oxid:
Hi, yes I am lacking of time, so I like to focus on more urgent pull requests.

PR #648, opened 2018-05-18
Remove  singleton  for instances of Base

mattvb91:

PR #657, opened 2018-06-28
Use registry session instead of problematic base singleton

mattvb91:
Same issue as with the config from https://github.com/OXID-eSales/oxideshop_ce/pull/648

Can someone from Oxid take a look at the tests as I think those tests were actually wrong as they were affected exactly by this singleton problem.


Sieg:
Hi @mattvb91, could you run and check these tests locally?:

`OxidEsales.EshopCommunity.Tests.Integration.Modules.ModuleNamespaceTest.testModuleWorksAfterActivation with data set #1
OxidEsales.EshopCommunity.Tests.Integration.Modules.ModuleNamespaceTest.testModuleDeactivation with data set #1
OxidEsales.EshopCommunity.Tests.Integration.Modules.ModuleNamespaceTest.testClassChainWithActivationAndDeactivation with data set #1
OxidEsales.EshopCommunity.Tests.Integration.Modules.ModuleNamespaceTest.testModuleActivationAndDeactivationUsesModulesMetadata with data set #1
OxidEsales.EshopCommunity.Tests.Integration.Modules.ModuleNamespaceTest.testModuleChainsGenerator_cleanModuleFromClassChain with data set #1
`

mattvb91:
@keywan-ghadami any chance you or someone on the oxid core team could take a look at the tests that understand them better? I cant see how based on this diff the tests would fail. It rather points to flakey tests based on the exact problem this PR fixes.

mattvb91:
@keywan-ghadami-oxid thats exactly whats happening. The copy made in Base::$_session gets out of sync with changes made through Registry::getSession()->bla() and vica versa if I remember correctly. 

So to my understanding the tests are actually using one of these out of sync session instances and this pull request has made this more obvious.

robertblank:
AFAIK we have the same problem potentially with those as well:

- \OxidEsales\EshopCommunity\Core\Base::$_blIsAdmin
- \OxidEsales\EshopCommunity\Core\Base::$_oActUser

with the caching, things may get out of sync and in such a case it is really hard to find the error.

Just imagine someone using it in production code and not just in tests ....
I would deprecate and later delete all those properties  and their setter and getter/methods and store the stuff in the registry as well ...

mattvb91:
@robertblank @keywan-ghadami-oxid any idea how to proceed with this?

I would see this rather as a bugfix instead of deprecating & bringing in on a major release?

keywan-ghadami-oxid:
@mattvb91 thank you very much for contributing, please have a look at comment from @Sieg , he is taking care of your PR but of cause the more you support him the better he can support on merging it.

If that topic is important to you but you do not have time to work on it, please consider to  create a bug ticket and/or contact product management. I will ask again in the public oxid dev slack.

mattvb91:
@keywan-ghadami-oxid I understand that but as I pointed out I think the failing tests are due to the bug discovered here and as a result im not sure how to approach fixing the tests as I think the assertions are based on faulty data to begin with.

Sieg:
Hello @mattvb91, @keywan-ghadami-oxid, so the "bug discovered here" is - using cached session instance as it should not be used? Maybe the bug can be reported with higher priority in bugtrack than? with information how to reproduce it, and than we could get more priority on this question! Sorry, i have not analized the tests, so cannot say whats wrong there, but its Integrations as i see, should be easy to run them and check.

robertblank:
I will try and have a quick look, why these tests fail

mattvb91:
> Hello @mattvb91, @keywan-ghadami-oxid, so the "bug discovered here" is - using cached session instance as it should not be used? Maybe the bug can be reported with higher priority in bugtrack than? with information how to reproduce it, and than we could get more priority on this question! Sorry, i have not analized the tests, so cannot say whats wrong there, but its Integrations as i see, should be easy to run them and check.

I think the easiest way to sum it up what is happening and @robertblank can correct me on this please is as follows: Every time a new object is created from Base it gets a copy of the session from registry as a singleton.

Now you have 100 objects with individual copies of session but if I make a change to anything inside registry::getSession() it will not be in sync with all of these copies which I think is whats happening with the tests (Now that they are all in sync because getSession() is the global registry object not copied for individual objects)

robertblank:
@mattvb91 you are perfectly right. The use of this class-level-cache is highly problematic, offers no benefit at all and was probably introduced just for testing purposes.

I think Base:getSession() is not so problematic.
It always should return a reference to the thing you get returned from Registry:getSession(). Changing any of these objects should sync all "copies".
AS LONG as you do not exchange the 'session' object in Base by something else by using Base:setSession(). For me Base:setSession() is the real evil guy. Then the "session" may get out of sync for some part of the code (and only god knows which!)

Another huge problem is, that the call stack of the first call differs from the call stack of the next calls, which might lead to hard-to-debug errors.

I really support this PR and with a fix in the testing library, the tests are also green:
https://travis-ci.org/OXID-eSales/oxideshop_ce/jobs/495501714

https://github.com/OXID-eSales/testing_library/compare/master-debug_core_base-mvb?expand=1
In this case it was really a difference in the call stack, which caused trouble - do not ask how long it took to figure that out :-(

IMHO This PR could even be backported to b-6.x - 
Furthermore we should really deprecate these methods and remove them in master

mattvb91:
wow nice find! I can definitely see the Base:setSession() being another issue. I would not have discovered that. I will resolve the merge conflict asap! Thanks for checking it out @robertblank

mattvb91:
Resolved the merge conflict. Please feel free to add any deprecated tags as you see fit.

mattvb91:
Pipeline green

PR #658, opened 2018-07-04
Bugfix/filecache race conditions

darchipov:
Simultaneous read/write operations on file caches lead to race conditions (missing or incomplete cache files)

-------------
file_get_contents(/XXXX/htdocs/tmp/config.1.adisabledmodules.txt): failed to open stream: No such file or directory
-------------
Parse error: syntax error, unexpected ''Dieser Online-Shop verwendet ' (T_ENCAPSED_AND_WHITESPACE) in /XXXX/htdocs/tmp/oxeec_langcache_0_0_1_premiumstore_default.txt on line 704

### DIFF
--- a/source/Core/FileCache.php
+++ b/source/Core/FileCache.php
@@ -53,7 +53,10 @@ class FileCache
     public function setToCache($key, $value)
     {
         $fileName = $this->getCacheFilePath($key);
-        file_put_contents($fileName, serialize($value), LOCK_EX);
+        $tmpFile = '/tmp/' . basename($fileName) . '.' . uniqid('', true) . '.tmp';
+        file_put_contents($tmpFile, serialize($value), LOCK_EX);
+
+        rename($tmpFile, $fileName);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


abejingas:
Are you suggesting, that exclusive lock doesn't work in php?

abejingas:
Are you suggesting that file locking doesn't work?

keywan-ghadami-oxid:
Hi I had the chance to talk to some developers on https://weuceu.org yes and there are some points that the rename solution is better:
- locking can lead to deadlock so having to deal with locking comes with additionally complexity that is against "KISS" 
- psr 6 and 16 comes without locking support in the interface so using file locking at this place would later be difficult to substitute with a standart. (and not all cache libraries support such locking, one I found that supporting such things is phpfastcache) 
- the renaming approach (that as stated before makes the file cache behave like any other cache), will allow future improvements using stale caches to improve performance, where the shared lock approach will allow to wait in any situation.

Sieg:
Hello @darchipov! Thank you for your improvement, i have merged it recently with some little changes on the top at e38295c6d70ab786cddfe910018a5cf66f6f3e37. Please let me know if i have missed something.

PR #659, opened 2018-07-04
Fixed return type in Basket::getDiscounts()

flow-control:
If there are no discounts in your basket, the `return null;` in this method leads to a

> E_WARNING: count(): Parameter must be an array or an object that implements Countable

in Basket::formatDiscount()


mattvb91:
Related to #652

Sieg:
Thank you @mattvb91 for your pull request! Merged to b-6.x recently (52e14e6e08ec8c741e922706441cdf9daac8f165)

PR #661, opened 2018-07-16
Deal with broken modules

keywan-ghadami-oxid:
With this commit the Oxid shop is able to handle broken modules in a proper way, by logging the exception without trying to disable the module. The old configurable behavior was eighter to throw a exception and by that stopping the shop from working or to disable the module, which could lead to endless loop or to unwanted behavior.
Todo

- [x] talk to architects 
- [x] fix code style
- [x] understand failing test / fix duplicate log entries
- [x] improve code comments
- [x] fix acceptance test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -100,6 +100,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Increase default quantity of productimages to 12 (from 7) [PR-514](https://github.com/OXID-eSales/oxideshop_ce/pull/514)
 - Make adding template blocks more fast andn reliable [PR-580](https://github.com/OXID-eSales/oxideshop_ce/pull/580)
 - Support PHP 7.2
+- Modules will not be disabled on class loading errors anymore, Error is just logged [PR-661](https://github.com/OXID-eSales/oxideshop_ce/pull/661)
 
 ### Removed
 - Removed old not used blAutoSearchOnCat option from shop_config tab [PR-654](https://github.com/OXID-eSales/oxideshop_ce/pull/654)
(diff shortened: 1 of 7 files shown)


keywan-ghadami-oxid:
@Sieg I finalized the pull request, If you like and when you have time we can have a talk about the changes.

There two changes that I feel I should mention them explicitly because they could be wrong:
1. In https://github.com/OXID-eSales/oxideshop_ce/pull/661/files#diff-1dc17aa3d190f03ef312b27e4e23bff9L346 I stopped class_exists from using the autoloader because it the autoloader would then again try to load the class and this would result in a second error in the logs. I am not sure if there was a reason why that "false" parameter was missing.
2. https://github.com/OXID-eSales/oxideshop_ce/pull/661/files#diff-bd16242248106cd839747d0205f7e50dL106 where I simplified the test to have only one class that will not be found.

mtkoltan:
I have talked with Keywan and in principle this change of behaviour makes sense. But I don't know if we need to consider this a BC break or not. I really would think that we could handle this as some sort of bug fix. I can't imagine that someone *relies* on the old behaviour for some purpose.

keywan-ghadami-oxid:
Some news on this, I was able to reproduce the issue by upgrading module internals. While fresh install works a upgrade e.g. from 1.0.1 to 1.2 makes the shop unusable.
Issue is also described here:
https://github.com/OXIDprojects/oxid-module-internals/issues/13

That module is good for testing this bug now because a file was renamed. In that case a "Module". It shows also that this pull request still have 2 edgecases to fix the module and the exception class are using created and may call the autoloader again and end up in a recursion. I am going to improve that code to catch that edge cases.

keywan-ghadami-oxid:
For the reason mentioned in the comment above I improved this pull request.

BenjaminJoerger:
It is fine to accept this PR as automatic disabling modules causes more issues tha
n  logging the exception. Like this we will give the shop owners the possibilty to control the shop by themselves.

Sieg:
@keywan-ghadami-oxid Some steps to improve were discussed with Keywan.

Sieg:
I have fixed those tests for you. Please double check if it makes sense: https://github.com/OXID-eSales/oxideshop_ce/commit/58607eb8a52d37893eed3a1d9a2ff99c66298988

keywan-ghadami-oxid:
yes thank you.

Sieg:
Hello @keywan-ghadami-oxid, merged to b-6.x and master recently.

PR #662, opened 2018-07-19
add missing trailing slash

keywan-ghadami-oxid:
without that slash modules with similar ids can get broken


Sieg:
Thank you @keywan-ghadami-oxid, merged to b-6.x recently (ef709f70a90eaa1dcf849ee72f1ab6e625ef56c3)

PR #666, opened 2018-09-13
Wrong return value FrontendController.isVatIncluded()

jctrant:
If an registered user is logged in without UStID, the FrontendController returns the wrong value for VAT included.
This is due to a type mismatch in PHP.
The if statement has a explicit type comparison to null, the given value from the database is an empty string ''. (Line: 2952 | Application/Controller/FrontendController.php)

Mantis #6902 (https://bugs.oxid-esales.com/view.php?id=6902)


Sieg:
Thanks for the fix @jctrant! I have merged it at 699a29f79e1daa8f8223353613e56d3dd2badad1

PR #668, opened 2018-10-01
Remove unused variables

Crease29:


source/Application/Controller/Admin/GenericImportMain.php
```diff
@@ -183,7 +183,7 @@ protected function _getCsvFieldsNames()
     {
         $blCsvContainsHeader = $this->getConfig()->getRequestParameter('blContainsHeader');
         \OxidEsales\Eshop\Core\Registry::getSession()->setVariable('blCsvContainsHeader', $blCsvContainsHeader);
-        $sCsvPath = $this->_getUploadedCsvFilePath();
+        $this->_getUploadedCsvFilePath();
```
Crease29: I'm not sure about this one so I only cleared the variable assignment. The called method seems to execute something relevant which is relevant for the current method.

source/Application/Controller/Admin/OrderMain.php
```diff
@@ -166,7 +166,7 @@ public function sendOrder()
             $oOrder->save();
 
             // #1071C
-            $oOrderArticles = $oOrder->getOrderArticles(true);
+            $oOrder->getOrderArticles(true);
```
Crease29: I'm not sure about this one so I only cleared the variable assignment. The called method seems to execute something relevant which is relevant for the current method.

Sieg:
Hello @Crease29, thank you for the improvements you did! I have rebased on latest b-6.x and merged it at 1459dafe753e1890924d71958bc8df353a407bf8 with some fixes from my side. Looks like some variables were still needed. Please double check if i have missed something and have a nice day! :)

Crease29:
Looks good to me! Thank you for rebasing and merging :slightly_smiling_face:

PR #670, opened 2018-10-12
added config file setting for database encoding

bmnnit:
Hi,

please review these small changes, the charset of the db connection is now configurable. Its not a full switch of the shop to utf8mb4 but ppl could decide for themself to convert the shop or single colums.

All Major CMS & Shopsystem support utf8mb4, emojies are an important SEO factor!

Johannes

### DIFF
--- a/source/Core/DatabaseProvider.php
+++ b/source/Core/DatabaseProvider.php
@@ -298,13 +298,15 @@ class DatabaseProvider
             ]
         ];
 
-        /** The charset has to be set during the connection to the database */
+        /** 
+         * The charset has to be set during the connection to the database
+         */
         $charset = (string) $this->getConfigParam('dbCharset');
-	//backwards compability with old config files
-        if ( null == $charset ) {
+        //backwards compability with old config files
+        if (null == $charset) {
             $charset = 'utf8';
         }
-
+        
         $connectionParameters['default'] = array_merge($connectionParameters['default'], ['connectionCharset' => $charset]);
 
         return $connectionParameters;
--- a/source/Core/DatabaseProvider.php
+++ b/source/Core/DatabaseProvider.php
@@ -299,7 +299,12 @@ class DatabaseProvider
         ];
 
         /** The charset has to be set during the connection to the database */
-        $charset = 'utf8';
+        $charset = (string) $this->getConfigParam('dbCharset');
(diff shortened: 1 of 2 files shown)


keywan-ghadami-oxid:
Hi Johannes, (I am not core developer, but maybe you like to help me to understand your pull request)

 utf8mb4 is sadly not the same as 'utf8' in MySQL syntax, so I personally agree with your request that using  utf8mb4 everywhere is a good thing and having emoji in search result is very cool. 

After having a look at you code I am wondering if it touches the right place because the value 'utf8' is already default. Can you put some lights on that and why there are no SQL to change the database?

bmnnit:
Hi Keywan,

for me this change would give me the possibility to change the database encoding just in the config.inc.php file. Right now i have to do a core change, and change Database Encoding there. With this patch its possible to set a different encoding (like utf8mb if wanted) but still keeps the old behavior. i havent done a full utf8m4 migration here, but opened the door to it, without a core change. So there are no sqls here because i didnt changed the encoding at all, but made it possible to do it. Its the first step to switch to utf8mb4.

keywan-ghadami-oxid:
Very cool, I like that, thank you.

Sieg:
Hello @bmnnit, thanks for the improvement! I do not see why this Backwards compatible feature should not be merged, so i have added some tests and merged at https://github.com/OXID-eSales/oxideshop_ce/commit/4d558466dea7ec1daaef906fe471febffdafab42. Please let me know if i have missed something.

ThomasDartsch:
Hello, 
there is still a special problem with the persparams. They are stored in oxorderarticles via serialize().
If there is a utf8mb4 character, the array becomes invalid and truncated in the db field.
With the result that the persparams for this order can not be read anymore.
TD

bmnnit:
i didnt include an alter db statements, since this is not what this commit is about. if you want to fully migrate oxid6 to utf8mb4 this is more work. my commit only makes it possible to set the encoding in the config file, which was only possible via are core hack before. please read my first comment on this. but where you are right, is that there should be a hint, that utf8mb4 is not fully supported yet.

ThomasDartsch:
I confirm that.
It was all about making it clear to oxid-developers that there are direct mistakes in using utf8.

PR #673, opened 2018-11-29
6886: exclude tests from dist-packages

kaluzki:


keywan-ghadami-oxid:
Note from a short look into: In the moment the commit is doing more then excluding files, it also has setting for git text handling

Sieg:
Hello guys. What you are suggesting is interesting stuff, and will be discussed with the team! But, anyway, to take this, we need to put some effort, as everything we have now: CI, parts of installation and development documentation are based on full dist packages, and used to test modules and shop itself.

For me personally, i had no problems to have those test directories nearby even on the live systems, so the question is for our PO - if we can spend time on this question, and if it's really beneficial?

Sieg:
Hello @kaluzki. Thanks a lot for your interest in improving the product! For this pull request, we will not merge it as it will introduce much more problems than benefits for our current infrastructure. The outcome of suggested changes is not hard to achieve - tests and other unwanted files can be removed during the deployment stage if they are really a problem.

PR #674, opened 2018-12-07
Bugfix for MultilanguageModel

tterhaarlaudert:
A bug occurs while editing a content page in the VCMS for the 8th language.
SEO-URL for oxcontent can't be updated if all other values (title, content, etc.) stay unchanged.
Because no values have to be updated in the table "oxcontents_set1" the function executeDatabaseQuery returns 0 for the amount of affected rows. In this case the update function would return false although the content has been saved successfully.

### DIFF
--- a/source/Core/Model/MultiLanguageModel.php
+++ b/source/Core/Model/MultiLanguageModel.php
@@ -463,7 +463,7 @@ class MultiLanguageModel extends \OxidEsales\Eshop\Core\Model\BaseModel
                 $insertSql = "insert into $langTable set " . $this->_getUpdateFieldsForTable($langTable, $this->getUseSkipSaveFields()) .
                              " on duplicate key update " . $this->_getUpdateFieldsForTable($langTable);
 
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
@reviewer I did a code review and
from code it seams to be correct. 

In general it is not correct to use return value of 'execute' to make assumptions about the success of the sql statement as it only returns the number of effected rows and that can be 0 if the value remains unchanged.

If there is really the need to know if there was a match in the database there is also the possibility to change that MySQL behavior to return always the amount of rows that match even if not changed by setting CLIENT_FOUND_ROWS  in the connection https://dev.mysql.com/doc/refman/8.0/en/mysql-affected-rows.html.

In the BaseModel _update will always return true independent of the number of affected rows so I guess that behavior would also be correct for this class.

Sieg:
Thank you @tterhaarlaudert for the fix, i have merged it recently to our b-6.1.x, so will be available with next patch. (https://github.com/OXID-eSales/oxideshop_ce/commit/3212ce1f352a7743c72669992077d488aef6663d)

PR #676, opened 2019-01-04
Code + PHPDoc improvements

KristianH:
set correct PHPDoc type's in Core/Controller/BaseController
added missing return statement in Core/Controller/BaseController
added missing property _aViewParams in Core/Controller/BaseController

### DIFF
--- a/source/Core/Controller/BaseController.php
+++ b/source/Core/Controller/BaseController.php
@@ -23,6 +23,13 @@ class BaseController extends \OxidEsales\Eshop\Core\Base
     protected $_aViewData = [];
 
     /**
+     * View parameters array
+     *
+     * @var array
+     */
+    protected $_aViewParams = [];
+
+    /**
      * Location of a executed class file.
      *
      * @var string
@@ -39,14 +46,14 @@ class BaseController extends \OxidEsales\Eshop\Core\Base
     /**
      * If this is a component we will have our parent view here.
      *
-     * @var \OxidEsales\Eshop\Core\Controller\BaseController
+     * @var \OxidEsales\Eshop\Core\Controller\BaseController|null
      */
     protected $_oParent = null;
 
     /**
      * Flag if this object is a component or not
      *
-     * @var bool
+     * @var bool|null
      */
     protected $_blIsComponent = false;
 
@@ -268,7 +275,7 @@ class BaseController extends \OxidEsales\Eshop\Core\Base
     /**
      * Sets value to parameter used by template engine.
      *
-     * @param string $sPara  name of parameter to pass
+     * @param mixed  $sPara  name of parameter to pass
      * @param string $sValue value of parameter
      */
     public function addTplParam($sPara, $sValue)
@@ -426,6 +433,8 @@ class BaseController extends \OxidEsales\Eshop\Core\Base
         if ($sParamId && isset($this->_aViewData[$sParamId])) {
             return $this->_aViewData[$sParamId];
         }
+
+        return null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


source/Core/Controller/BaseController.php
```diff
@@ -268,7 +275,7 @@ public function addGlobalParams($oShop = null)
     /**
      * Sets value to parameter used by template engine.
      *
-     * @param string $sPara  name of parameter to pass
+     * @param mixed  $sPara  name of parameter to pass
```
keywan-ghadami-oxid: In that line above I guess the function parameter type `sting` in the comment was correct for the parameter name but needs to be changed for the value

source/Core/Controller/BaseController.php
```diff
@@ -867,6 +875,8 @@ public function getSidForWidget()
         if (!$oSession->isActualSidInCookie()) {
             return $oSession->getId();
         }
+
+        return null;
```
keywan-ghadami-oxid: in the line above: you added "return null;" which I would not recommend for php developers to do explicitly as this is the default and adds no value. For perl or other languages it's another story.

source/Core/Controller/BaseController.php
KristianH: corrected in https://github.com/OXID-eSales/oxideshop_ce/pull/676/commits/5eab81d2860e99f5db83cd83e1afc94c1e78550d

source/Core/Controller/BaseController.php
KristianH: solved in https://github.com/OXID-eSales/oxideshop_ce/pull/676/commits/dc7fde207e69ebd01aff5227f7482aa48771b68e and https://github.com/OXID-eSales/oxideshop_ce/pull/676/commits/4e2501d1fa6d46697c12df0114a4047cead6787e

mattvb91:
@KristianH I've been working on a small lib that could automagically generate PHPDoc with associated types for all off oxids models from the DB structure: https://github.com/mattvb91/php-doc-prop-generator

On the "formatter" branch theres a wip for an oxid formatter. Could potentially included that to document all models along with giving an upgrade path for PHP 7.4 types. Any interest if I make a pull request with it to cover the core oxid models for a start?

source/Core/Controller/BaseController.php
```diff
@@ -423,9 +430,12 @@ public function getViewData()
      */
     public function getViewDataElement($sParamId = null)
     {
+        $return = null;
         if ($sParamId && isset($this->_aViewData[$sParamId])) {
-            return $this->_aViewData[$sParamId];
+            $return = $this->_aViewData[$sParamId];
         }
+
+        return $return;
```
alfredbez: I don't understand, what you're trying to do here, since the return value is already `null` if you don't return any other value:
```
Psy Shell v0.9.9 (PHP 7.2.10-0ubuntu0.18.04.1 — cli) by Justin Hileman
>>> function foo() { }
>>> function bar() { return; }
>>> function baz() { return null; }
>>> foo();
=> null
>>> bar();
=> null
>>> baz();
=> null
```

Same goes for the changes in `public function getSidForWidget()`

KristianH: i got a inspection information on my IDE:
Missing return statement.
Inspection info:
The inspection is intended to point at the following types of inconsistencies:
- The function/method contains return statements with arguments as well as without them
- The function/method may happen to return a value or end its execution without a return statement at all
- The function/method is declared void but contains return statements with arguments

Technically these are not errors but practically usually indicate programmer’s mistakes.

Should I remove the changes in both methods?

KristianH:
> Any interest if I make a pull request with it to cover the core oxid models for a start?
@mattvb91  Sure why not, sounds interesting for me.

PR #677, opened 2019-01-09
remove return in construct-method

alfredbez:
`__construct` cannot return a value


Sieg:
Hello @alfredbez, thanks for improvement. Merged to b-6.x and master

PR #681, opened 2019-02-05
Fix: ini_set allowed shown as red

Sioweb:
Try to set memory_limit with its same value, is a non invasive method to check if ini_set is possible. Dont see a reason to config a fully setup session.

Tested with Oxid PE 6.1.2

### DIFF
--- a/source/Core/SystemRequirements.php
+++ b/source/Core/SystemRequirements.php
@@ -741,7 +741,7 @@ class SystemRequirements
      */
     public function checkIniSet()
     {
-        return (@ini_set('session.name', 'sid') !== false) ? 2 : 0;
+        return (@ini_set('memory_limit', @ini_get('memory_limit')) !== false) ? 2 : 0;
     }
 
     /**


keywan-ghadami-oxid:
@Sioweb hi I'am trying to understand the need of that change. If you like please give me some insides why you improved that special place (e.g. was it blocking the installation, or just show in the admin area), why was it red for setting "session.name" but green for changing the memory?

Sioweb:
Hi this fixes only the system requirements check. I am not sure why exactly session.name not work - or any other session config - but i think its because the session is already in use. I red on issue tracker that some ppl has this problem and i've choosen a setting which has no effect on runtime. 

Well one thing could harm this code, if ini_get is disabled and ini_set not, maybe this should be checked first in this code. Otherwise i dont know what will be set as memory limit.

Sieg:
Hello @Sioweb, you mentioned there are some bugs on bugtracker. Could you please post some links to those bugs? As for now, i dont see any reasons to change the original code.

Sioweb:
1.) https://bugs.oxid-esales.com/view.php?id=6849
2.) https://forum.oxid-esales.com/t/4-10-8-php-7-2-ini-set-rot-und-internal-server-errors-auf-einigen-seiten/94554
3.) https://github.com/zendframework/zend-session/issues/104

Well could be caused in PHP >= v7.2 because its not allowed anymore to change session values after `session_start`. 

You could try it here: https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/Session.php#L330

        echo (ini_set('session.name', 'asd') != false) ? 2 : 0;
        $this->_blStarted = @session_start();
        echo (ini_set('session.name', 'asd') != false) ? 2 : 0;
        die;

Result: 20

Sieg:
Thank you @Sioweb, i have merged your suggested change to b-6.x(70d06531798f469867afa721ae7f1934ec602d3d) and up to master.

PR #683, opened 2019-02-07
ensure temp file in tmp directory

keywan-ghadami-oxid:
with this fix temporary files are ensured to be created within the tmp directory
the issue without the fix is that in a secure envuornment and without a trailing slash in compiledir folder setting  the confic.inc.php the temp file can not be created and for example orders failed.
a workaround for projects is to lower security for writing next to the tmp folder or to add a trailing slash in the setting for the tmp folder.
no bugtracker entry created yet by me


Sieg:
Hello @keywan-ghadami-oxid, merged to b-6.1.x, will be available in next patch

PR #685, opened 2019-02-08
make shop forks work out of box

kaluzki:
Hello,

for my private projects I use a fork of the oxid shop (oxidio/shop).
This package is registered on packagist.org and will be copied to "vendor/oxidio/shop" when installed with composer.

To make this constellation executable, I had to patch two places so far:

* https://github.com/oxidio/shop/commit/52e8189c3ac05a119da57f8d57380ee4a60f752d
* https://github.com/oxidio/facts/commit/1f56d43deec740c0ab7c66b374a14cee746428ba

My previous installations were based on db dumps. 
Yesterday, for the first time, I tried to build an environment from scratch.
That did not work, which is the reason for this PR.

For the current adaptation, I presuppose the class `\OxidEsales\Facts\Facts`.
In this context, the question arises:

Why the package `oxide-esales/oxideshop-unified-namespace-generator` and thus also `oxide-esales/oxideshop-facts` are not defined in the `require` composer section?
The `Fact` dependencies occur directly though only in the `Setup` namespace, but over the dependency to the generated classes in
`vendor/oxide-esales/oxideshop-unified-namespace-generator/generated/` the dependencies are also given at runtime.

Basically, I want a solution that allows the forks to run out of box.

### DIFF
--- a/source/Setup/functions.php
+++ b/source/Setup/functions.php
@@ -48,10 +48,11 @@ if (!function_exists('getCountryList')) {
      */
     function getCountryList()
     {
+        $cePath = (new Facts)->getCommunityEditionSourcePath();
         $aCountries = [];
         $relativePath = 'Application/Controller/Admin/ShopCountries.php';
-        if (file_exists(getVendorDirectory() . "oxid-esales/oxideshop-ce/source/$relativePath")) {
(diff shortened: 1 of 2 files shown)


Sieg:
Hi @kaluzki, thanks for the improvement, i have merged it to b-6.x and up. Additionally, that condition is not needed anymore, as facts have it inside already.

PR #686, opened 2019-02-13
Fixed exception after delete of delivery-address

JaroslavHerber:
If user deletes a selected delivery-address and then goes to basket -> exception.
The reason is the not deleted session-var "deladrid", which is used to load delivery-address in Order->getDelAddressInfo()

### DIFF
--- a/source/Application/Model/Order.php
+++ b/source/Application/Model/Order.php
@@ -1148,7 +1148,10 @@ class Order extends \OxidEsales\Eshop\Core\Model\BaseModel
         }
         if ($soxAddressId) {
             $oDelAdress = oxNew(\OxidEsales\Eshop\Application\Model\Address::class);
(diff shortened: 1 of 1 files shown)


JaroslavHerber:
Sorry for the double-pulls. The file-change off pull request #687 should be part of this one too.

Sieg:
Hello @JaroslavHerber, please check the reply on #687. I have problems on reproducing the problem, please describe more. Maybe there are some additional modules configured? Please double check.

Sieg:
@JaroslavHerber Would be great if you will update one of the pull requests to be Complete, dont want to play with it. You can simply add new commits to the pull request branch, and it should be updated automatically.

Sieg:
Hello @JaroslavHerber, closing this one as well, as i cannot reproduce the issue, and there are no more responses from your side :(

PR #688, opened 2019-02-18
Add blocks to order_info.tpl

gaertnermarkus:
Add two blocks to order_info.tpl.
Util now it is not possible to add more rows inside the table.


Sieg:
Hi @gaertnermarkus, have merged your request to b-6.x, so should be available in 6.2.0 compilation. Have a nice day.

PR #689, opened 2019-02-22
Remove the file xd_receiver.htm

michaelkeiluweit:
The feature Facebook Integration came with the version 4.4.0 in 2010: https://oxidforge.org/en/new-features-in-oxid-eshop-version-4-4.html
Since then the file xd_reciever.htm is an ingredient of the shop packages and as of today it is a relic of the Facebook Integration.

The file basically calls a script from the URL http://static.ak.connect.facebook.com/js/api_lib/v0.4/XdCommReceiver.js. The URL isn't callable anymore and so I guess no one uses the included JavaScript file since years. I also didn't find any replacement for it. As a result of this it should be removable easily.

This PR adds a migration script to remove the file from older installations during a migration and it removes the file from the repository.

### DIFF
--- a/source/migration/data/Version20190222120038.php
+++ b/source/migration/data/Version20190222120038.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace OxidEsales\EshopCommunity\Migrations;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+use OxidEsales\Eshop\Core\Utils;
+use OxidEsales\Eshop\Core\UtilsFile;
+
+/**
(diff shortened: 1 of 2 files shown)


Sieg:
Hey @michaelkeiluweit. Thanks for the pull request!

We had some discussions about this question recently in the team, and that's what we have decided so far:

* Removing files with **database migration scripts** looks not a very clean solution
* The idea of removing files automatically is complicated as well, as someone might still use them.

So, as outcome, i am merging half of your pull request - the part where we actually remove the file, and additionally we have added the section in shop update procedure, where those old unused files will be listed, so developers could remove them by hand, if they don't see any use of those.

PR #690, opened 2019-02-22
Fix: Run-time warning in inc_error.tpl

stephaneisenbach:
Warning: count(): Parameter must be an array or an object that implements Countable in /source/tmp/smarty/9f82ed8fbbbe090e6a26df004c8d6113^%%47^47B^47B0D5FE%%inc_error.tpl.php on line 4

See screenshot.

bildschirmfoto vom 2019-02-23 00-03-03


Sieg:
Hello @stephaneisenbach, thanks for the fix, merged to b-6.1.x, so will be in next patch.

PR #691, opened 2019-02-25
Small improvement of the description of the method assign

michaelkeiluweit:


Sieg:
Hello @michaelkeiluweit, that method always return null, isnt it? Please reopen the Pull request if i did.

michaelkeiluweit:
@Sieg Yes it always returns null. But the description says it would maybe return "true".

PR #692, opened 2019-02-28
The tpl language is accidently used for saving the base language...

michaelkeiluweit:
...which causes some broken view sql queries later.

Please see https://bugs.oxid-esales.com/view.php?id=6957#c12802 (List item #2)


Sieg:
Thanks @michaelkeiluweit, merged to b-6.1.x and up

PR #693, opened 2019-03-04
namespaced model constants for better development experience

kaluzki:
* quick documentation
* show usages
* show type declaration
* completion
* refactoring
* ...


keywan-ghadami-oxid:
Hi @kaluzki, I think using constants for table names is a perfect change. thank you for your work on this.

For fields and columns I am wondering if they sharing the same underlining information. I think having them both increases the amount of code to be maintained and processed and therefor should be avoided. So maybe field name constants can be dropped from your Pullrequest (and later, 
If you and the core team agrees (because I am not working on the oxid core, and this is only my personal thougths) for code using field names, that code can be changed to use column names and getters.

best regards

Sieg:
Hello @kaluzki, we've had some discussion in team regarding your suggestion in this pull request. So, first of all, thank you for the ideas you provide - we really appreciate it!

Regarding this one, we can't take it into the core, as it is actually not what we want:
* "Hardcoded" model names are on their place now already in models, so there is nothing wrong about it. Maybe it could switch to constant, but not to separate namespace - the constant for the model.
* Regarding the names of the fields: we would like to have getters and setters in models to handle fields in the future, so there will be no need to have those field names separately somewhere.
* And last, but not least - if we will take something like this, we cannot revert this decision later.

Any of described questions needs much more discussions in the team, as we trying to not expand our API with new methods for now, for the sake of improving the infrastructure and make it better from what we already have. Hope everyone will like the new stuff coming up with OXID 6.2.

PR #697, opened 2019-03-23
Changed the default email encoding to 'base64'

dx-bhesse:
Prevents issues caused by emails containing long lines in some email-clients like Outlook
(Outlook sometimes decodes emails with 'quoted-printable'-encoding incorrectly if long lines are involved)
Side effect of base64-encoding: the email size will increase slightly

Related: https://bugs.oxid-esales.com/view.php?id=6468


Sieg:
Hello @dx-bhesse, tested, looks good, merged to b-6.x and master. Thanks for the improvement!

PR #698, opened 2019-04-08
Read config parameter by getConfigParam, not by getRequestParam.

michaelkeiluweit:
By calling getRequestParam('blVariantSelection') we will always get null, as this request is never made and leads to this bug: https://bugs.oxid-esales.com/view.php?id=6968


Sieg:
Hi @michaelkeiluweit, yah, i have fixed that one already. Merged to b-6.1.x (https://github.com/OXID-eSales/oxideshop_ce/commit/c68b9a0e77f927b9686c1fa67153967cce95b701). Thanks!

PR #700, opened 2019-04-11
test against all php7 versions

alfredbez:
I think it makes sense to run travis builds for all supported and even some future php versions, since you can allow failures in travis.


alfredbez:
It's basically the same as #518 but for newer php versions.

Sieg:
Looks like shop is not working with at least php 7.3 yet, so will not add it to travis till we can have green light there. Not talking about 7.4 at all yet :)

alfredbez:
That's exactly why I added these versions to `allow_failures`

Sieg:
Hey @alfredbez, thanks for the PR. Have just merged to b-6.x(7dd7828cef69cf0bfdf249235290a60e70964127) and up.

PR #702, opened 2019-05-06
add setStockStatus()

proudcommerce:


Sieg:
Hello @proudcommerce, could you please describe more, why we need to set this flag manually?

tabsl:
Changing this flag through a module?! ;-)[@proudcommerce]

Sieg:
So if we need to change it in the module, its possible to extend the article as well, in the module, and add the method there. Are there any use cases it should be in the article class itself?

tabsl:
I can’t say it exactly anymore, but we needed it ...

Sieg:
No, you don't.

Sieg:
Please reopen the pull request if there will be a list of use cases for this method, otherwise, we don't see any point to add it - as mentioned already, this flag is calculated on loading and stock recalculation actions, and by the idea, should not be changed manually.

PR #704, opened 2019-05-14
Fix finalize order doesn't validate payment with user from parameter

zaxxo:
Some modules finalizes order without set user globally.
Instead they pass user by parameter to Order->finalizeOrder function.
This function calls Order->isValidPayment function.
But here the globally set user is used and it can happen that payment is not valid.


zaxxo:
Closed, wrong pull request branch.

PR #707, opened 2019-05-23
bugfix 6985: fix sql error in category sort ajax popup

alfredbez:
fixes https://bugs.oxid-esales.com/view.php?id=6985

Short screencast of the bug:
OXID-Bug-category-sort


Sieg:
Hello @alfredbez, merged to b-6.x and up to master. Thanks!

PR #708, opened 2019-05-27
removed EXCEPTION_LOG, logs are now in oxideshop.log

ivoba:
Removes all occurences of EXCEPTION_LOG in favour of oxideshop.log.

Not sure if occiurrence in .htaccess and robots.txt should be left for legacy support?


Sieg:
Hello @ivoba, merged to b-6.x and up to master. Thanks!

PR #710, opened 2019-06-12
adapted code style  to PSR-2

alfredbez:
I've just executed `php-cs-fixer fix . --rules=@PSR2` and checked the diff, looks like there are some weird changes, especially for comments.


Sieg:
Good job there, but so sad its based not on b-6.x :D we could have it earlier!

alfredbez:
Take a look at #711

PR #711, opened 2019-06-12
applied automatic fixes via php-cs-fixer and phpcbf

alfredbez:
see #710


Sieg:
Thanks @alfredbez, merged to b-6.x

PR #712, opened 2019-07-01
bugfix 6671

Pauleo77:
Using the amount prices in the backend when recalculating the basket.
Currently the amount prices are just working in the frontend (basket).

Bugfix for the following issue:
https://bugs.oxid-esales.com/view.php?id=6671


Sieg:
Hey @Pauleo77! Thanks a lof for the fix! Looks good! Have merged it to b-6.1.x and up.

PR #713, opened 2019-07-08
Fix error message bug

C3000:
Steps to repreduce:
1. Open the shop
2. Add an article to the basket
3. Go to checkout and login as user
4. Change the adress so that an error comes up (testing: disable JS)
5. Submit the form again with error

Expected behavior:
There is one error in the top of the page (value expected)

Actual behavior:
There are two errors in the top of the page (value expected, value expected)

P.S.
Normally there is no double error message because the errors will be cleared by the widgets which will be called after.
If there aren't any widgets the bug will be visible.

### DIFF
--- a/source/Application/Component/UserComponent.php
+++ b/source/Application/Component/UserComponent.php
@@ -726,7 +726,6 @@ class UserComponent extends \OxidEsales\Eshop\Core\Controller\BaseController
             return;
         } catch (\OxidEsales\Eshop\Core\Exception\InputException $oEx) {
             Registry::getUtilsView()->addErrorToDisplay($oEx, false, true);
-            Registry::getUtilsView()->addErrorToDisplay($oEx, false, true, 'input_not_all_fields');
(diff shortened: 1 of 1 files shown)


Sieg:
Hello @C3000, Thanks for the pull request. The problem is - i cannot reproduce the error. Tested our latest frow theme with b-6.x Enterprise and Community edition shops. Could you please double check what is your configuration? Maybe you are using modified error display or some other then flow theme?

C3000:
Hello @Sieg ,
yes, we're not using the flow theme. The difference between our theme and the flow theme is that we don't use widgets. This should be an allowed behaviour.
In the flow theme there is no double error message because the errors will be cleared by the widgets which are called after.
If there aren't any widgets the bug will appear.
Regards

Sieg:
Hey @C3000, thanks for the fix. I have rebased it on b-6.3.x and merged up.

PR #714, opened 2019-07-16
Catch PHP Fatal error.

tmloberon:
If a wrong string is entered in the paramerter `listtype=`, a PHP Fatal Error will occur. You will be thrown back to the home page instead of displaying the detail page.

Triggered with uri: /Kiteboarding/Kites/Kite-CORE-GTS.html?listtype=XYZ_BOOM


Sieg:
Thanks @tmloberon! Merged.

PR #716, opened 2019-07-23
remove unnecessary parameters in function call

alfredbez:
see https://github.com/OXID-eSales/oxideshop_ce/blob/96dcc008e9d0a4643fb8363e07c359682751faeb/source/Core/UtilsView.php#L139-L146

`$blFull` is already set to false and `$useCustomDestination` makes only sense if you provide also `$customDestination` (which we don't do in this case).


Sieg:
Thanks @alfredbez, merged!.

PR #718, opened 2019-07-29
Pass fetchAll parameters to get more control

Sioweb:
PDO::fetchAll has some arguments, which can control the output. If we use `SELECT OXID FROM oxarticles;`, we would get 

```
Array
(
    [0] => Array
        (
            [0] => 022ec39af56268f3cd232d69494b8fc5
        )

    [1] => Array
        (
            [0] => 0292f9ba6f4dffff4e4c4bf66328163a
        )

    [2] => Array
        (
            [0] => 03022a216baaee246aa6790572264309
        )
....
```

But with `->fetchAll(\PDO::FETCH_COLUMN)` we would get:

```
Array
(
    [0] => 022ec39af56268f3cd232d69494b8fc5
    [1] => 0292f9ba6f4dffff4e4c4bf66328163a
    [2] => 03022a216baaee246aa6790572264309
    ....
```

This would help to make some modules much more performent, because ppl do not need to walk every row in a while-loop or use array_walk or something.

What do u think about it?

### DIFF
--- a/source/Core/Database/Adapter/Doctrine/ResultSet.php
+++ b/source/Core/Database/Adapter/Doctrine/ResultSet.php
@@ -84,14 +84,17 @@ class ResultSet implements \IteratorAggregate, ResultSetInterface
     /**
      * Returns an array containing all of the result set rows
      *
+     * @param integer $iFetchArguments controlls how PDO should format result array.
+     * @param integer $sFetchClass will initiate a class for each row if needed.
+     *
      * @return array
      */
-    public function fetchAll()
+    public function fetchAll($iFetchArguments = null, $sFetchClass = null)
     {
         $this->close();
         $this->getStatement()->execute();
 
-        return $this->getStatement()->fetchAll();
+        return $this->getStatement()->fetchAll($iFetchArguments, $iFetchClass);
     }
 
     /**


Sioweb:
it would also be totaly okay to use something like:

```php
$args = func_get_args();
call_user_func_array([$this->getStatement(), 'fetchAll'], $args);
```

Just to be sure all parameters will be passed as PHP it intended.

flow-control:
Hey @Sioweb,

i personaly think this is great idea, but in my opinion we should try to match the upstream method signature from `Doctrine\DBAL\Driver\ResultStatement::fetchAll()`.
https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Driver/ResultStatement.php#L87

What do you think?

Kind regards
Florian

Sioweb:
Ho!

im not sure if this will break compatibility. The first argument of doctrines fetchAll is fetchMode. What happen to all modules which still use fetchAll() with arguments as first parameter?

Or did i missed something?

flow-control:
Hi,

could you clearify that?
AFAIK the `fetchAll()` method never had any arguments.

Kind regards
Flo

Sioweb:
Ah sure! 

i've just checked my changes and not the original code :)

Well im not sure if this methode is using doctrine, i thought it use PDO? If not, i agree, we can use doctrines parameters.

flow-control:
Hi,

OXID is using Doctrine which is using PDO to connect to MySQL. Therefore - as a side note - you should not use `\PDO::FETCH_COLUMN`, but `\Doctrine\DBAL\FetchMode::COLUMN`.

/Flo

Sioweb:
Alright, ill update the PR then?

flow-control:
From my point of view: yes :+1: 
@Sieg, what do you think about this?

Sieg:
Hello guys, first of all - thanks for your interest in improving the OXID. 

For this concrete PR, I think that this will be not very good solution, as:
* we will not follow the interface with this implementation anymore.. Yes, it will work for this case, but if the ruleset will be extended somewhere, the extended class will not work anymore.
* it should be simple to use, and thats, why we have simple getAll and **getCol** for your case
* adding strange integer arguments to switch the results, is not a good tone. If you really need some other type of result, better option in my opinion would be to add some specific method for it, or just build the whole chain to get your needed result at that place.

I will move this PR to our Architects board, so they could give a green light for it if i am wrong with my points.

Sieg:
Hello. Looks like architects agree with my points regarding this one.

PR #720, opened 2019-08-01
bugfix for https://bugs.oxid-esales.com/view.php?id=6811

Alpha-Sys:
First part of bugfix for https://bugs.oxid-esales.com/view.php?id=6811
Use correct return value.


Sieg:
Thanks @Alpha-Sys, merged. The bug is not yet closed as frontend should be fixed in flow and wave.

PR #722, opened 2019-09-11
SEO Thing ... URI and end slash 

TumTum:
If the end slash is missing at the URI, then it is clearly a 301 forward and not a temporary forward.

```
     http://oxid.localhost/Kiteboarding/Trapeze  
  -> HTTP-Status 301 
  -> http://oxid.localhost/Kiteboarding/Trapeze/

```


Sieg:
Hello @TumTum, thanks for the improvement! Looks correct, so merged to our b-6.x and master.

PR #723, opened 2019-09-27
Improved default e-mail validator expression

SvenBrunk:
no longer accepted:
sven.brunk@-.-,-,-.oxid-esales.com
sven.brunk@-esales.com

### DIFF
--- a/source/Core/MailValidator.php
+++ b/source/Core/MailValidator.php
@@ -24,7 +24,7 @@ class MailValidator
     public function getMailValidationRule()
     {
         if (is_null($this->_sMailValidationRule)) {
-            $this->_sMailValidationRule = "/^([\w+\-.])+\@([\w\-.])+\.([A-Za-z]{2,64})$/i";
+            $this->_sMailValidationRule = "/^(\w+([\.\-\+]){0,1})+\@(\w+([\.\-]){0,1})+([A-Za-z]{2,64})$/i";
         }
 
         return $this->_sMailValidationRule;


SvenBrunk:
Remark: I escaped some more characters than needed: within [] escaping + and . is actually not needed

SvenBrunk:
nice! Github also does it wrong 🗡 🤣

Sieg:
Hi @SvenBrunk, thanks for the improvement! So as i understood, the main point was to add the "+" symbol possible, right?

What could be improved in the suggested version:

- "{0,1}" part could go as just "?"
- we are missing the "\." part from original expression now, was it bad? the email like "itis@working" is fine with your variant.

I would also suggest considering usage of native PHP functionality as default filter - filter_var('bob@example.com', FILTER_VALIDATE_EMAIL), what do you think?

flow-control:
Hey @Sieg,

@SvenBrunk and i discussed that, and we came to the conclusion, that switching to `filter_var` would be the best solution, but not that easily doable when we do not want to to a bc break, as the API would change in that case.
We could change it to use `filter_var` for the next major release though.

/Flo

flow-control:
Additional Info: we can remove the `getMailValidationRule`, `setMailValidationRule` and the `__construct` Methods in that case. I would totally go for it, but that would be a BC Break ;-)

flow-control:
I removed the reg ex completely in #725, so we could merge this into the current dev and merge the other on in master, what do you think @Sieg ?

Sieg:
@flow-control, @SvenBrunk Hey guys! What about just extracting the Default case and run it if we are not setting new rules?, so, lets say, we have rule as null, and if its null, we are using the default filter_var for checking the email, otherwise, we use what Developer wants. That will be BC compatible, and will use filter_var by default?

flow-control:
I would be ok with that @Sieg, but in the long run i would prefer to remove that altogether with the next major release. If someone still want's to check the email validity with it's own regular expression, the `MailValidator` class can be overloaded.

flow-control:
Hey @SvenBrunk,
@Sieg and i had a discussion about this and we think that we could push this even further towards the `filter_var` thing. Could you add that as a default in the `isValidEmail` Method by checking if `self::__sMailValidationRule` is still `null` and if so, call `filter_var` and if not, use the regex.

/Flo

Sieg:
Hello, Thank you for your ideas! After some internal discussions, there are EmailValidatorService and EmailValidatorServiceBridge introduced. Please use those instead of old stuff!

PR #724, opened 2019-10-03
Make backend "save in language" compatible with lowercase fields

zsolt-szilagyi:
Make Backend "Save in Language" compatible with lowercase and CamelCase fields.

Previously oxid was not able to provide the "save in lang" dropdown in the backend of custom models if these had lowercase or camelcase fieldnames (e.g. mytable__title instead of mytable__TITLE). The reason was that, while searching for used fields, the fieldnames got uppercased.  Since all of the shop is fine with lowercase fields, it makes sense to remove that final barrier.


Sieg:
Hello @zsolt-szilagyi! Thanks for your improvements! It looks like we have 3 condition blocks with the same body, so would be better to combine that condition, how do you think?
If it is possible, please describe your use case in more detail, what are you trying to do, how to use those multilanguage fields, and what cases are not working and working before and after the change. It will be much easier for us to investigate the suggested change with such information.

PR #725, opened 2019-10-03
switched from regular expression to filter_var for email validation

flow-control:
This pull request removes the regular expression for validating emails completely, but this is a bc break
related to #723


source/Core/InputValidator.php
```diff
@@ -186,14 +186,26 @@ public function checkEmail($user, $email)
...
+    /**
+     * @param strgin $email
+     * @return bool
+     */
+    public function isValidEmail($email)
+    {
+        if (!is_string($email) || !strlen($email)) {
+            return false;
+        }
+        return filter_var($email, FILTER_VALIDATE_EMAIL) !== false;
```
alfredbez: Maybe use this instead:
```php
return filter_var($email, FILTER_VALIDATE_EMAIL) 
  && preg_match('/@.+\./', $email);
```
... and add a test for `abc@def` as well

tests/Unit/Core/InputValidatorTest.php
```diff
@@ -1197,4 +1197,43 @@ public function testGetCompanyVatInValidator_DefaultTurnedOffOnline()
...
+            ['mytest@oxid-esales.museum', true],
+            ['?mathias.krieck@oxid-esales.com', true],
+            ['my/test@com.org', true],
+            ['mytest@-com.org', false],
+            ['@com.org', false],
+            ['mytestcom.org', false],
+            ['foo.bar@-.-,-,-.oxid-esales.com', false],
+            ['mytest@com', false],
+            ['info@�vyturys.lt', false],
+        ];
```
alfredbez: You could add `abc@def` to the list, it should pass now, but I guess that's not an expected mail address.

source/Core/Autoload/UnifiedNameSpaceClassMap.php
```diff
@@ -2853,12 +2853,6 @@
         'isInterface'      => false,
         'isDeprecated'     => false
     ],
-    'OxidEsales\Eshop\Core\MailValidator'                                        => [
```
Sieg: We could deprecate it (isDeprecated -> false), but not remove the section, i guess.

flow-control: This class does not exist anymore in the pull request, but we could deprecate it in #723, what do you think? #723 should target `b-6.x`, so we can deprecate the class there and then finally remove it with this pull request that targets `master`

tests/Unit/Core/InputValidatorTest.php
flow-control: Hey @alfredbez,
with `mytest@com` we have a email address with a similar schem in the data provider which evaluates to `false` -> not a valid email address. In my personal opinion `abc@def` should also be not accepted.
/Flo

source/Core/InputValidator.php
flow-control: see https://github.com/OXID-eSales/oxideshop_ce/pull/725#discussion_r335406376

tests/Unit/Core/InputValidatorTest.php
alfredbez: oh, I missed that :man_facepalming:

source/Core/InputValidator.php
alfredbez: Ok, I rechecked that and that shouldn't be a problem since PHP 5.3.3 (which is really old...)
image

PR #726, opened 2019-10-08
Removed "http://" from value of oxarticles__oxexturl input.

Phillinger:
Hardcoded prefix "http://" has been removed from theme **flow** and **wave**. 

Discussion in forum: [https://forum.oxid-esales.com/t/externe-url-und-https/94781](https://forum.oxid-esales.com/t/externe-url-und-https/94781)

Prefix in admin template makes no sense anymore. You save "https://example.com", after reload there's "http://https://example.com" in input field.


Sieg:
Hello @Phillinger, thanks for the fix. Additionally to this, i have removed the manipulation with "http://" from the controller, so it should be fine now. Merged to b-6.x and master.

PR #728, opened 2019-10-10
do not mess with error_reporting server setting

flow-control:
This pull requests removes resetting the PHP error reporting level via a call to `error_reporting()` in the `ShopControl` class.

I personally think it is not a good idea to mess with server settings in such a way, that it just overrides whatever the server settings are. In most cases the operator put thought in these settings and we should not mess with these.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         "oxid-esales/oxideshop-ide-helper": "dev-master",
         "oxid-esales/azure-theme": "dev-master",
         "oxid-esales/oxideshop-facts": "dev-master",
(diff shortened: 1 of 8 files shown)


flow-control:
@Sieg do you think we can target `b-6.x` with this PR?

flow-control:
Rebased on latest `master`

flow-control:
Rebased on lastest `master` and pushed

flow-control:
Depends on https://github.com/OXID-eSales/testing_library/pull/23 being merged upfront

Sieg:
Hey @flow-control, this one is merged! :) Have a nice day!

PR #729, opened 2019-10-22
remove string concatenations in help lang files

kermie:
Especially in help_lang files for admin panel, string concatenations are used for better readability, which is definitely a good thing. OXID handles that natively but other systems that have to work with these files don't by default, for example standard translation software.
Of course, I've been talking to @BenjaminJoerger about that and we agreed that it would be better to sacrifice readability for interoperability in this case.
Additionally, I removed unnecessary line breaks as well as changed a language value in EN when escaping a single quotation mark, replacing double quotes of this language string with single ones.


Sieg:
Merged to b-6.x and master.

PR #730, opened 2019-10-25
cleaned mixed up left and right and switched to english comment

SvenBrunk:
switched names of the blocks as the left one was named right and the right one named left


SvenBrunk:
image

Sieg:
Hello @SvenBrunk, thank you for the fix. Merged to our master, so will be available only with the next major version.

PR #731, opened 2019-10-31
BUGFIX: since the exception's message is already translated in InputV…

mreiss:
…alidator, we need to switch/case the translation itself (and not the ident). Maybe refactor to proper exception-handling some day.


Sieg:
Hi @mreiss, is there a bug reported on the bug tracker regarding this case? Could you please link it if there is one?

Sieg:
Hello @mreiss, thanks for the fix. Merged to b-6.2.x and up, so will be available in the next 6.2 patch.

PR #732, opened 2019-11-01
Add logging to shop constructor if shop is not valid

alfredbez:
We had a broken license key, which resulted in some strange SQL-Errors like this:
```
An exception occurred while executing 'select .`oxid` from  where 1  and .oxid = '2''
```
After a good amount of time we figured out, that the license key was expired and the shop was not valid, `$this->init('oxshops');` was not called and that was the reason for the missing table name in the above SQL-Statement. It would be helpful to see that in the logs I guess.


Sieg:
Resolved with #733.

PR #733, opened 2019-11-01
add logging to shop constructor if shop is not valid

alfredbez:
see also #732


Sieg:
Hello @alfredbez. Thanks for the improvement. Merged this to b-6.x (4efe146fd7cdc2b7572018d1ee8dadd8f55b1519) and up to master. Will close another one as this already covers the functionality from #732.

PR #734, opened 2019-11-06
Fix query command detection

GM-Alex:
The query command detection fails for the following query:

```MySQL
SELECT
1 FROM test
```

The result of `getFirstCommandInStatement` is `SELECT 1` instead of `SELECT`, see http://sandbox.onlinephpfunctions.com/code/7d64d49e511e2b3729a82287bc9f7e75cba15df9 for an example. The reason is that the regex doesn't detect the line break after the SELECT with `.*` even if the `s` modifier is set. To fix this we first replace all line breaks with a space. So `getOne` works for all queries instead of returning `false`.

### DIFF
--- a/source/Core/Database/Adapter/Doctrine/Database.php
+++ b/source/Core/Database/Adapter/Doctrine/Database.php
@@ -1330,8 +1330,9 @@ class Database implements DatabaseInterface
      */
     protected function getFirstCommandInStatement($query)
     {
+        $singleLineQuery = str_replace(["\r", "\n"], ' ', $query);
         $sqlComments = '@(([\'"]).*?[^\\\]\2)|((?:\#|--).*?$|/\*(?:[^/*]|/(?!\*)|\*(?!/)|(?R))*\*\/)\s*|(?<=;)\s+@ms';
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
without this fix shop will not execute the given sql so the newly developed code (modules) will not work when using this kind of sql.

Sieg:
Hello, but this will replace all new lines in the query, isnt that too much? :/ If you will try working with some text, it will change the content. Sure, prepared statements should be used, but still?

keywan-ghadami-oxid:
what about using 
```
preg_split('/ |\n/', trim($uncommentedQuery))[0]
```
(a example can be found here http://sandbox.onlinephpfunctions.com/code/138397f1ba6f6fef4a3bbe71259b638a6c4eee93) in that example code I also tried to simplify the regex but I do not yet suggest to use that unless there are some test to prove things.

GM-Alex:
Since the code (including the manipulation of the query) is only used to determine the type (select, update etc.) of the query, should it be safe to replace the new lines.

Sieg:
Hello @GM-Alex, thanks for the improvement! Merged to b-6.1.x and up, so will be available with the next patch for 6.1 and up.

PR #736, opened 2019-11-14
Simplification, the notation, to get quickly the QueryBuilder

TumTum:
Code should be easy and quick to read. I would suggest to call it as usual via the Registry::class.

This is how it looks now in v6.5.0:

```php
\OxidEsales\EshopCommunity\Internal\Container\ContainerFactory::getInstance()
    ->getContainer()
    ->get(\OxidEsales\EshopCommunity\Internal\Framework\Database\QueryBuilderFactoryInterface::class)
    ->create()
        ->select('oxarticels')
        ->where('oxid = :oxid')
        ->setParameter('oxid', 'a1b2b3')
        ->execute();
```

and it could look like this:

```php
Registry::getQueryBuilder()
    ->create()
        ->select('oxarticels')
        ->where('oxid = :oxid')
        ->setParameter('oxid', 'a1b2b3')
        ->execute();
```

### DIFF
--- a/source/Core/DatabaseProvider.php
+++ b/source/Core/DatabaseProvider.php
@@ -13,7 +13,7 @@ use OxidEsales\Eshop\Core\Exception\DatabaseNotConfiguredException;
 /**
  * Database connection class
  *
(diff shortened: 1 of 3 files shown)


Sieg:
Depends on #737 decision.

PR #737, opened 2019-11-14
Simplification, the notation, to get quickly the DI container.

TumTum:
Code should be easy and quick to read. I would suggest to call it as usual via the Registry::class.

This is how it looks now in v6.5.0:

```php
\OxidEsales\EshopCommunity\Internal\Container\ContainerFactory::getInstance()
    ->getContainer()
    ->get(\OxidEsales\EshopCommunity\Internal\SomeInterface::class)
```

and it could look like this:

```php
Registry::getContainer()->get(\OxidEsales\EshopCommunity\Internal\SomeInterface::class)
```

### DIFF
--- a/source/Core/Registry.php
+++ b/source/Core/Registry.php
@@ -328,6 +328,22 @@ class Registry
     }
 
     /**
+     * Return \Symfony\Component\DependencyInjection\Container
+     *
+     * @return \Psr\Container\ContainerInterface
+     */
+    public static function getContainer()
+    {
+        $class = \Psr\Container\ContainerInterface::class;
+
+        if (!isset(self::$instances[$class])) {
+            self::$instances[$class] = \OxidEsales\EshopCommunity\Internal\Application\ContainerFactory::getInstance()->getContainer();
+        }
+
+        return self::$instances[$class];
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


Sieg:
Hello @TumTum, we have the getContainer method in the base class: https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/Base.php#L147

TumTum:
Hi @Sieg  that means that I must inherit all my business domain classes from Base. However, there is situations where i don't want to have the overhead of base.

I like to write my business code outside the Oxid Framework. To not be dependent on the OXID.  However, I would be happy if I could call the container via the Registry::class or as a global function.

iegupov:
Hi @TumTum if you want to use our services, why do not register your class as a service in yaml file and inject our services in the constructor of your class? The ContainerFactory mainly was meant to be used in traditional classes which extend the Base class.

Sieg:
Hello @TumTum, so, is there anything wrong with injecting the stuff via constructor?

Sieg:
Hello, as there are no communication about this question anymore, i am closing it. Methods mentioned by Igor should be used to make stuff simple and easily readable.

PR #738, opened 2019-11-18
Fixed sorting for cross-selling article lists

tterhaarlaudert:
Added "order by" clause to cross-selling query


Sieg:
Hi @tterhaarlaudert, thanks for the improvement. Merged.

PR #739, opened 2019-11-20
0007051: added resetFilter

ivoba:
Fix for https://bugs.oxid-esales.com/view.php?id=7051

When resetting Filter on ArticleList an error occurs:
ERROR_MESSAGE_SYSTEMCOMPONENT_FUNCTIONNOTFOUND resetFilter ["[object] (OxidEsales\\Eshop\\Core\\Exception\\SystemComponentException(code: 0):

### DIFF
--- a/source/Application/Controller/ArticleListController.php
+++ b/source/Application/Controller/ArticleListController.php
@@ -346,6 +346,17 @@ class ArticleListController extends \OxidEsales\Eshop\Application\Controller\Fro
     }
 
     /**
+     * Reset filter.
+     */
+    public function resetFilter()
+    {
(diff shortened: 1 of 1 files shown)


ivoba:
@Sieg could you pls check this.
This is a acknowledged bug and i would like to upgrade my shops without forking, pls.

Sieg:
Hello @ivoba. Thanks for pointing out the problem. I have merged the change to b-6.1.x and up just to not break the updated theme. As you already said, the method already exists in 6.2 shop branch, but the theme have no 6.2 shop requirements, and you got the newer theme with functionality which is not yet released for the shop:( We will review this issue and will be more careful about it in the future.

PR #740, opened 2019-11-22
Bug 0007014: Second page of article list doesn't load via SEO

dx-bhesse:
https://bugs.oxid-esales.com/view.php?id=7014

convertSeoPageStringToActualPageNumber() called in extractPageNumberFromSeoUrl() already decreases the pgNr-parameter by 1
.../1/ will set $pageNumber=0 but will not get set as $urlParameters['pgNr'], 1<0 is false
.../2/ will set $pageNumber=1 but will not get set as $urlParameters['pgNr'], 1<1 is false <- this is the bug-case, the page-number gets lost
.../3/ will set $pageNumber=2 and will get set as $urlParameters['pgNr'], 1<2 is true <- starts working from here on

Often oxid-shops append the pgNr-parameter to the seo-urls and do not encode them directly into the seo-url and are therefore not affected by this bug.


dx-bhesse:
$data['old_style_paginated_page'] = ['params'   => $this->seoUrl . '2/',
                                             'expected' => ['cl'   => 'alist',
                                                            'cnid' => $this->categoryOxid,
                                                            'lang' => '0']
        ];

...not quite sure if the test is missing the pgNr-key or if those seo-urls with paginations in them were simply deprecated some time ago and should not longer be used?

dx-bhesse:
> not quite sure if the test is missing the pgNr-key 

I think the pgNr-Key was missing in the test. But still not quite sure if this is pull-request now is what the oxid functionality intended to todo

Sieg:
Hello @dx-bhesse, thanks for the pull request! Looks correct. Merged.

PR #742, opened 2019-11-27
Added check if discounts variable is an array

zoh48gz04:
Similiar to PR-678, fixes warnings in PHP >= 7.2


Sieg:
Hello @zoh48gz04, thanks for the improvement. Merged to b-6.2.x so will be available with 6.2 patch

PR #745, opened 2019-12-04
Migrate from PHP 7.4snapshot to 7.4 in travis

flow-control:
As PHP `7.4` is now official released, Travis upgraded their images too and PHP `7.4` can now be used without the `snapshot` suffix


Sieg:
Hey @flow-control, looks good. Merged.

PR #746, opened 2019-12-05
Bump DI version due to  CVE-2019-10910

flow-control:


alfredbez:
Composer will use the latest version of `v3.4` anyway or am I wrong? See also: https://semver.mwl.be/#!?package=symfony%2Fdependency-injection&version=%5E3.4&minimum-stability=stable

flow-control:
That's not granted, another dependency could require an explicit version and as we allow it, it would install. Better save then sorry :wink:

flow-control:
Additionally one could also use the `--prefer-lowest` argument with `composer install` and composer will install the lowest allowed version.

alfredbez:
I think adding `roave/security-advisories` is also a good idea:
```
composer require --dev roave/security-advisories:dev-master
```
see: https://github.com/Roave/SecurityAdvisories

Sieg:
Hey @flow-control, merged to b-6.2.x and up.

PR #747, opened 2019-12-05
changed _getDbConfigTypeName from private to protected

moritzdemmer:
changing private to protected makes it really easy to add new settings (for example a colorpicker) in the metadata.

Extending ModuleConfiguration like that, adds a new type "colorpicker" to the metadata.php.

```
public function init()
    {
        $this->_aConfParams["colorpicker"]    = 'confclr';
        parent::init(); // TODO: Change the autogenerated stub
    }

protected function getDbConfigTypeName($type)
    {
        parent::getDbConfigTypeName($type);
        return $type === 'colorpicker' ? 'str' : $type;
    }

```

using block "admin_module_config_var_types" in module_config.tpl allows us to add the new type to the template
```
[{$smarty.block.parent}]
[{if $var_type == 'colorpicker'}]
<input type="color"  name="confclr[[{$module_var}]]" value="[{$confclr.$module_var}]" >
 [{/if}]
```

In the metadata.php you can add the colorpicker with
```
'settings' => array(
                  array('group' =>  'color_main', 'name'  =>  'color_picker', 'type'  =>  'colorpicker', 'value' =>  '#ff9b00')
    )
```

### DIFF
--- a/source/Application/Controller/Admin/ModuleConfiguration.php
+++ b/source/Application/Controller/Admin/ModuleConfiguration.php
@@ -334,7 +334,6 @@ class ModuleConfiguration extends \OxidEsales\Eshop\Application\Controller\Admin
      */
     protected function getDbConfigTypeName($type)
     {
-       parent::getDbConfigTypeName($type);
-       return $type === 'colorpicker' ? 'str' : $type;
+        return $type === 'password' ? 'str' : $type;
     }
 }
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
2 cents on this, maybe better more like this:
```
 protected function getDbConfigTypeName($sType)
 {
      $sType= parent::getDbConfigTypeName();
      return $sType === 'colorpicker' ? 'str' : $sType;
  }
```
- directly without underscore "_" because now it is no BC break to fix that
- just for reference/documentation: always use parent call to avoid conflicts with other modules

Sieg:
Hi @moritzdemmer, thanks for your interest in improving the system! 

Looks like this pull request is not viable anymore, as the method _loadMetadataConfVars is deprecated and not used anywhere anymore. Take a look at Internal\Framework\Module package.

PR #748, opened 2019-12-10
Feature fix type annotations

flow-control:
While working on https://github.com/OXID-eSales/graphql-example-module/pull/5 I saw that the type hints for `$shopId` are `string` instead of `int` in some occasions. 

This pull request
- changes the type hints from `string` to `int` where they are wrong for `$shopId` argument
- changes naming from `$shopID` to `$shopId` to be consistent throughout the source

This is no BC break


Sieg:
Hey @flow-control, thanks for the improvement. Merged to b-6.x and up to master.

PR #749, opened 2019-12-19
OXDEV-2892 - Replace our container builder with lcobucci/di-builder

flow-control:


flow-control:
Team decision was to not use the di builder

PR #750, opened 2019-12-23
Skip url generation if "Display Currencies" option is disabled

tterhaarlaudert:
This change solves performance problems on article list pages (ArticleListController, SearchController). For every ArticleBox-Widget on the page the CurrencyComponent will be loaded. If the "Display Currencies" option is disabled we don't the URL that will be generated in the components render function. (Tested with OXID EE 6.1.0 + Memcached + Varnish Cache)


Sieg:
Hello @tterhaarlaudert, thanks for the improvement! Merged to master.

PR #751, opened 2020-01-02
fix wrong return type in NavigationController@_doStartUpChecks

alfredbez:


Sieg:
Hello @alfredbez, thanks for the fix! Merged to b-6.2.x and up to master.

PR #754, opened 2020-01-09
fix view generation on broken shop

keywan-ghadami-oxid:
this happens if there is a table in the configuration 
that do not exists. ignoring this issue here allows the user to generate the views and get a running shop.

without this fix there is sql error in that case because there are no fields to build a valid sql

### DIFF
--- a/source/Application/Model/Shop.php
+++ b/source/Application/Model/Shop.php
@@ -343,6 +343,13 @@ class Shop extends \OxidEsales\Eshop\Core\Model\MultiLanguageModel
             $sJoin = $this->_getViewJoinLang($table, $languageId);
         }
 
+        if ("" === $sFields) {
+            //this happens if there is a table in the configuration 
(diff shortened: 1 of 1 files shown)


source/Application/Model/Shop.php
```diff
@@ -343,6 +343,13 @@ protected function addViewLanguageQuery($queryStart, $table, $languageId, $langu
             $sJoin = $this->_getViewJoinLang($table, $languageId);
         }
 
+        if ("" === $sFields) {
+            //this happens if there is a table in the configuration
```
Sieg: Lets log this case as warning maybe?

keywan-ghadami-oxid: yes, good point, I am going to add this

Sieg:
Merged to b-6.x(ae49fca4eda39673669fbc12cc08766ea327f44e) and up to master.

PR #755, opened 2020-01-10
Added possibility to add a custom offline-page

JaroslavHerber:
It should be possible to overwrite the static 500-status offline.html
This standard-page and status-code isn't good for SEO at all.

### DIFF
--- a/source/bootstrap.php
+++ b/source/bootstrap.php
@@ -228,8 +228,7 @@ function oxTriggerOfflinePageDisplay()
 
         /**
          * Render an error message.
-         * If offline.php exists its content is displayed.
-         * Like this the error message is overridable within that file.
+         * If offline.html exists its content is displayed.
          */
         if (is_readable(OX_OFFLINE_FILE)) {
             echo file_get_contents(OX_OFFLINE_FILE);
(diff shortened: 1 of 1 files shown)


Sieg:
Hello @JaroslavHerber! Thanks for the improvement! We have added this question to discuss during our plannings, and we will have in mind your implementation. There are a lot of questions around those offline pages for now.

Sieg:
Hey @JaroslavHerber, How about - we will add the check if the "oxTriggerOfflinePageDisplay" is defined, so you could write your own?

Loesdau:
Hey @Sieg, so you mean you would move oxTriggerOfflinePageDisplay to overridablefunctions.php with function_exists() check? And this function could be overwritten in /modules/functions.php for example?
This sounds great too :)

JaroslavHerber:
Sorry for the last comment from another account. Thats was me 😅

Sieg:
Hello @JaroslavHerber, I have merged what I was talking about :) (https://github.com/OXID-eSales/oxideshop_ce/commit/bf39ed6276f40291932cd523bf3bff631d9e2ad9) Hope that will help. Have a nice day!

PR #761, opened 2020-02-04
fix code style with phpcbf

alfredbez:
There were [86 violations](https://travis-ci.org/OXID-eSales/oxideshop_ce/jobs/645524000#L1111) which can be fixed automatically.

I've executed this to fix those:
```
vendor/bin/phpcbf --standard=PSR12 --extensions=php source/
```

During adding the changed files I saw also a typo and a too-long line which I fixed alongside.

The remaining [18 violations](https://travis-ci.org/alfredbez/oxideshop_ce/jobs/645851634#L1115) which can be fixed automatically are in the azure theme.

PR #762, opened 2020-02-04
fix code style with phpcbf

alfredbez:
see #761


Sieg:
Hello @alfredbez, thanks for the effort, but it looks like after our recent internal workflow changes, this PR is not required anymore, we will have master as our current development branch.

PR #767, opened 2020-02-21
do not throw an exception if the correct controller is already present

alfredbez:
I don't know the exact steps to reproduce this, but I already saw this a couple of times:

image

Sometimes the controllers are already in the database (previously configured?) and are correct, so I think we should only complain if there is really a problematic scenario, where we try to map two different classes to one controller name.

I'm not sure if this is related to https://github.com/OXIDprojects/oxid-module-internals/issues/103

### DIFF
--- a/source/Internal/Framework/Module/Setup/Validator/ControllersValidator.php
+++ b/source/Internal/Framework/Module/Setup/Validator/ControllersValidator.php
@@ -142,7 +142,13 @@ class ControllersValidator implements ModuleConfigurationValidatorInterface
         $controllerClassNameSpaces = [];
 
         foreach ($controllers as $controller) {
-            $controllerClassNameSpaces[] = $controller->getControllerClassNameSpace();
+            $controllerClassname = $controller->getControllerClassNameSpace();
+            if (
+                !isset($controllerClassMap[$controller->getId()])
+                || $controllerClassMap[$controller->getId()] !== $controllerClassname
+            ) {
+                $controllerClassNameSpaces[] = $controller->getControllerClassNameSpace();
+            }
         }
 
         return array_intersect($controllerClassNameSpaces, $controllerClassMap);
--- a/tests/Unit/Internal/Framework/Module/Setup/Validator/ControllersModuleSettingValidatorTest.php
+++ b/tests/Unit/Internal/Framework/Module/Setup/Validator/ControllersModuleSettingValidatorTest.php
@@ -12,10 +12,15 @@ namespace OxidEsales\EshopCommunity\Tests\Unit\Internal\Framework\Module\Setup\V
 use OxidEsales\EshopCommunity\Internal\Framework\Config\Dao\ShopConfigurationSettingDaoInterface;
 use OxidEsales\EshopCommunity\Internal\Framework\Config\DataObject\ShopConfigurationSetting;
 use OxidEsales\EshopCommunity\Internal\Transition\Adapter\ShopAdapterInterface;
-use OxidEsales\EshopCommunity\Internal\Framework\Module\Configuration\DataObject\ModuleConfiguration;
-use OxidEsales\EshopCommunity\Internal\Framework\Module\Setup\Validator\ControllersValidator;
+use OxidEsales\EshopCommunity\Internal\Framework\Module\Setup\{
+    Exception\ControllersDuplicationModuleConfigurationException,
+    Validator\ControllersValidator,
+};
+use OxidEsales\EshopCommunity\Internal\Framework\Module\Configuration\DataObject\{
+    ModuleConfiguration,
+    ModuleConfiguration\Controller,
+};
 use PHPUnit\Framework\TestCase;
-use OxidEsales\EshopCommunity\Internal\Framework\Module\Configuration\DataObject\ModuleConfiguration\Controller;
 
 /**
  * @internal
@@ -61,15 +66,21 @@ class ControllersModuleSettingValidatorTest extends TestCase
     }
 
     /**
-     * @expectedException \OxidEsales\EshopCommunity\Internal\Framework\Module\Setup\Exception\ControllersDuplicationModuleConfigurationException
-     *
      * @dataProvider duplicatedSettingValueDataProvider
      *
      * @param Controller[] $duplicatedSettingValue
+     * @param bool $expectException
      *
      */
-    public function testValidationWithDuplicatedControllerNamespace(array $duplicatedSettingValue)
-    {
+    public function testValidationWithDuplicatedControllerNamespace(
+        array $duplicatedSettingValue,
+        bool $expectException
+    ) {
+        if ($expectException) {
+            $this->expectException(
+                ControllersDuplicationModuleConfigurationException::class
+            );
+        }
         $shopAdapter = $this->getMockBuilder(ShopAdapterInterface::class)->getMock();
         $shopAdapter
             ->method('getShopControllerClassMap')
@@ -104,19 +115,28 @@ class ControllersModuleSettingValidatorTest extends TestCase
         }
 
         $validator->validate($moduleConfiguration, 1);
+
+        if (!$expectException) {
+            $this->assertTrue(true);
+        }
     }
 
     public function duplicatedSettingValueDataProvider(): array
     {
         return [
-            [
+            'same controller names with different classname' => [
                 [
                     new Controller('moduleControllerName', 'duplicatedNamespace'),
-                ],
-                [
                     new Controller('duplicatedName', 'moduleControllerNamepace'),
                 ],
-            ]
+                'expectException?' => true
+            ],
+            'same controller names with same classname' => [
+                [
+                    new Controller('duplicatedname', 'duplicatedNamespace'),
+                ],
+                'expectException?' => false
+            ],
         ];
     }
 }


IgorIe:
This scenario shouldn't happen. Could you provide steps how to reproduce it? We validate controllers only on the module activation and before the module activation controllers must not present in the database. So it just silences the problem.

In case if multiple modules have the same controller id there won't be an error and developers will search for a very funny bug trying to understand why some module controller doesn't work anymore.

alfredbez:
> This scenario shouldn't happen. Could you provide steps how to reproduce it?

I usually do, but I don't have an idea, what exactly causes this problem. I understand that this shouldn't happen, but it happens anyway :-)

If I find a way how to reproduce this, I will definitely let you know.

---

> In case if multiple modules have the same controller id there won't be an error and developers will search for a very funny bug trying to understand why some module controller doesn't work anymore.

Nope, it will still throw an exception in this scenario, let me explain this real quick:

Image we have this controller map in the shop already:
```
'foocontroller' => 'Vendor\Module\Controller\FooController'
```

Now we try to activate a module with this controller-map:
```
'foocontroller' => 'Vendor\Module\Controller\FooController'
```

:heavy_check_mark: That's fine, since it's the exact same mapping, but if we now have another module which tries to do this:

```
'foocontroller' => 'OtherVendor\OtherModule\Controller\FooController'
```

... we should definitely throw an exception as we already do. I'm just talking about the part where we already have the correct controller map in the db and we're trying to insert it again.

alfredbez:
See also:

https://github.com/OXID-eSales/oxideshop_ce/pull/767/files#diff-e86cc35f1abfc1f37e410c08268c2c48R88

https://github.com/OXID-eSales/oxideshop_ce/pull/767/files#diff-e86cc35f1abfc1f37e410c08268c2c48R136

keywan-ghadami-oxid:
This is one of the most scaring things that stops me using oxid 6.2 rc for some time now.
I am also not sure how this can happen, but as I saw this happening in the past I would love to see the shop being robust against this situation. 

Maybe it is related to tools messing around with the internal state (like module internals) or to broken deactivation or activation of modules (e.g. during module development). Anyway I think the shop should should be able to deal with that situation at least in a way that it allows the user to bring the system back into defined state. 

@alfredbez I like you pull request, so I going to add a cat picture for better acceptance.
grafik

IgorIe:
If you get this exception I pretty sure that you have completely inconsistent modules data in your database, and it's not only about controllers, but about class extensions, templates, blocks etc.

It can happen, for example, if you update from 6.1 to 6.2 without doing mentioned in the update documentation steps (see https://docs.oxid-esales.com/developer/en/6.2-rc.1/update/index.html#update-of-the-module-configurations ), or if you switch between 6.1 and 6.2 and have shared database, or use developer tools which are not compatible with version 6.2 (for example, they write module data to the database, it won't work this way anymore).

If you have this state you may get many problems different problems (not active modules will work, removed module extensions could be included in the chain etc) and it's very good that there is a validation and we can spot this issue. Removing validation won't solve the problem.

You can run several console commands from the update documentation and get a proper state according to your modules configuration in yml file:

```
vendor/bin/oe-console oe:oxideshop-update-component:delete-module-data-from-database
vendor/bin/oe-console oe:module:apply-configuration
```
It should fix your database data, but anyway, we have to figure out why it happens and solve it in a proper way.

PS: A nice try with cat!

keywan-ghadami-oxid:
@IgorIe thank you for the information and new way of thinking about this. I agree with you that we need a warning in this case and find the root cause. On the other hand this validation stops the admin from activating a module and that is a no go for onlineshops that may need to activate a module for some reason. The oxid shop needs to robust in case of error conditions and admin user must be able to control it (without command line)
So there is not only black and white you can of cause have warnings and error messages in such situation and the system to continue working. 

I think all this validation is good thing in general, and also beeing strict to keep things consistens is a good thing. Data redundancy should be avoided and this principle was not used for modulr state, sometimes more then three location (metadata/database/filesystem/caches) are used, and the need to keep things consistent and the complexity is becomming challenging. Because of this it will never be a easy solution to guarantie robustness and keeping the code clean unless that tedundancy will be removed.

So I vote for:
- first: making the shop robust against issues with the internal state and keeping the error messages
- second: removing all duplication of state into the database and files that are not only for caching or deployment

IgorIe:
@keywan-ghadami-oxid thank you for your ideas. For the first topic I would like to understand how it happens and where, is it really a case for administrator of online shop. All cases which I know for now are related to switching between versions or doing update without needed steps, development cases. If there are other cases, would be good to check them.

For the second topic - totally agree, for example there's no much sense to have this data in the database, except of backward compatibility. We have this topic on our desk and will think what we can do with it.

alfredbez:
In our case, it happened in a new project where we started from scratch with a 6.2 installation, so no upgrade, etc. We did not stumble upon this during module development, but simply by using the Visual CMS module. I heard from other devs, that they encountered this problem during unit-test execution.

Sioweb:
In my case this error occurs, if a module cant find a class or something and it deactivates itself and i try to reactivate it.

Maybe modules are not correctly deactivated in case of an error?

IgorIe:
Hey guys,

we refactored controllers validator and changed a bit the logic from the pull request:
https://github.com/OXID-eSales/oxideshop_ce/commit/95b8754a29c0cd0246819fea92b2919ce5e608a3#diff-c5beb93cc2088655b5cf7679e5196154R69

Exception won't be thrown, but an error will be logged, module activation isn't blocked for user anymore. Sadly we still didn't get any real case how to reproduce the issue.

Thank you for your help!

alfredbez:
FYI: Many developers asked me if I know this error, here are a few cases (copied from the chat, some of them are in german):

> Christopher Bauer Jan 22nd at 11:12 AM
> Hallo :slightly_smiling_face:
> ist jemanden von euch ein Problem bei den OXID UnitTests bekannt, das er nach dem UnitTest die Module nicht korrekt in der DB freigibt?
> "Fehler: Modul nicht aktiviert.<br />In einem Shop oder Sub-Shop müssen Schüssel und Werte für "controllers" eindeutig sein.<br />Folgende Schlüssel oder Werte für "controllers", die in der Datei metadata.php dieses Moduls definiert sind, existieren schon im Shop:<br />"

> Stefan Moises  4 months ago
> Ich hatte den Fehler in einer EE auch schon ohne UnitTests … einfach so bei Aktivierungen in mehreren Subshops

> Vivien Schwarz  4 months ago
> Hatte den selben Fehler, ich konnte das Reproduzieren, als ich das Modul bereits registriert hatte und Änderungen daran vorgenommen hatte (namespaces o.ä.), danach wurde es mir deaktiviert angezeigt, konnte aber nicht mehr aktiviert werden.
> Ich musste das Modul komplett rauskicken, im Backend die Moduleinträge dann rauschmeißen und danach wieder installieren. Anders ging es für mich nicht zu lösen.
> Ist auch nicht wirklich schön :confused:
> War hier auch ne neue 6.1.5 EE mit 3 Subshops.

> Stefan Moises Mar 19th at 9:22 AM
> Gibt es eigentlich für diesen Fehler auch einen Fix in der 6.2. EE? “In einem Shop oder Sub-Shop müssen Schüssel und Werte für “controllers” eindeutig sein.”

> Sascha Weidner  2 months ago
> Das passiert bei mir immer, wenn ein Modul sich selbst deaktiviert, weil ne Klasse o.Ä. nicht gefunden werden kann und User das Modul mit Controller wieder aktiviert.

> Oliver Anderson  11:58 Uhr
> And this is not consistent, on the first run the tests pass, but on the second run I get this error
> ```
> Adding unit tests from /var/www/oxideshop/source/modules/oxps/sepa/Tests//unit/modules/application/models/*Test\.php
> PHPUnit 4.8.26 by Sebastian Bergmann and contributors.
> .......
> Time: 1.88 seconds, Memory: 16.00MB
> OK (7 tests, 17 assertions)
> ```
> Then on the second run:
> ```
> Adding unit tests from /var/www/oxideshop/source/modules/oxps/sepa/Tests//unit/modules/application/models/*Test\.php
> PHP Fatal error:  Uncaught OxidEsales\Eshop\Core\Exception\StandardException: Fehler: Modul nicht aktiviert.<br />In einem Shop oder Sub-Shop müssen Schüssel und Werte für "controllers" eindeutig sein.<br />Folgende Schlüssel oder Werte für "controllers", die in der Datei metadata.php dieses Moduls definiert sind, existieren schon im Shop:<br /> OxidProfessionalServices\Sepa\Application\Controller\Admin\SepaController,OxidProfessionalServices\Sepa\Application\Controller\OxpsSepaAgreement in /var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/UtilsObject.php:231
> Stack trace:
> #0 /var/www/oxideshop/source/oxfunctions.php(101): OxidEsales\EshopCommunity\Core\UtilsObject->oxNew('OxidEsales\\Esho...')
> #1 /var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/Module/ModuleInstaller.php(108): oxNew('OxidEsales\\Esho...')
> #2 /var/www/oxideshop/vendor/oxid-esales/testing-library/library/ModuleLoader.php(88): OxidEsales\EshopCommunity\Core\Module\ModuleInstaller->activate(Object(OxidEsales\Eshop\Core\Module\Module in /var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/UtilsObject.php on line 231
> PHP Warning:  Cannot modify header information - headers already sent by (output started at /var/www/oxideshop/vendor/oxid-esales/testing-library/AllTestsRunner.php:7) in /var/www/oxideshop/source/bootstrap.php on line 74
> PHP Stack trace:
> PHP   1. OxidEsales\TestingLibrary\Bootstrap\UnitBootstrap->{closure:/var/www/oxideshop/source/bootstrap.php:41-78}() /var/www/oxideshop/source/bootstrap.php:0
> PHP   2. setcookie() /var/www/oxideshop/source/bootstrap.php:74
> PHP Warning:  Cannot modify header information - headers already sent by (output started at /var/www/oxideshop/vendor/oxid-esales/testing-library/AllTestsRunner.php:7) in /var/www/oxideshop/source/bootstrap.php on line 75
> PHP Stack trace:
> PHP   1. OxidEsales\TestingLibrary\Bootstrap\UnitBootstrap->{closure:/var/www/oxideshop/source/bootstrap.php:41-78}() /var/www/oxideshop/source/bootstrap.php:0
> PHP   2. setcookie() /var/www/oxideshop/source/bootstrap.php:75
> ```

> 12:01 Uhr
> I don't think those values have been set

> Marc Plotz Today um 09:00
> Has anyone seen something like this before, and why is it suddenly complaining about this when nothing has changed?
> ```
> [2020-05-28 08:31:20] OXID Logger.ERROR: Module data discrepancy error: module data (controller with id amazonconfig and namespace: OxidProfessionalServices\AmazonPay\Controller\Admin\ConfigController ) for module oxps/amazonpay was present in the database before the module activation [] []
> ```

> Marc Plotz  vor 17 Minuten
> More context - I had to rebuild my containers in the template project this morning, so the db dump is taken from the same file that’s been in use for quite a while.

Sioweb:
Yesterday it happend to me, with oxid 6.1.

1. Install module with controller
2. Activate module
3. Add a new class in metadata => "extend"
4. Reactivate module
5. Duplicated controller error occurs
6. Activate module again

Just for completeness, dont now if it is still important or not, with the new controller validation.

alfredbez:
Maybe we should also consider a similar patch in OXID 6.1

PR #769, opened 2020-02-28
Cleanup PHPCodeSniffer usage and reported errors

flow-control:
Hello there :wave:,

this pull request:

- integrates PHPCodeSniffer the official way :heavy_check_mark: 
- provides a `phpcs.xml.dist` :heavy_check_mark: (configuration in a file and not hidden in the code of another repository) 
- cleans up with all the errors reported from PHP CodeSniffer :heavy_check_mark: 
- enforces PSR12 coding guidelines :heavy_check_mark: 
- allows us to deprecate [OXID-eSales/coding_standards_wrapper](https://github.com/OXID-eSales/coding_standards_wrapper) :fire:
- leaves all the warnings from PHPCodeSniffer untouched, so we still have a lot of work to do :wink:
- runs PHPCodeSniffer in Travis without the warnings :tada: 

If merged, this allows for the use of the PHP Code Beautifier and Fixer at `vendor/bin/phpcbf` that is shipped with the PHP CodeSniffer to just fix the coding style errors.

/Flo

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -7,9 +7,7 @@
     <arg name="colors"/>
     <arg value="p"/>
     <arg value="n"/>
-    <arg name="parallel" value="75" />
+    <arg name="parallel" value="75"/>
     <arg name="extensions" value="php"/>
-    <rule ref="PSR12">
-        <exclude name="PSR2.Methods.MethodDeclaration.Underscore"/>
-    </rule>
+    <rule ref="PSR12"/>
 </ruleset>
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -1,0 +1,15 @@
+<?xml version="1.0"?>
+<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="PHP_CodeSniffer" xsi:noNamespaceSchemaLocation="phpcs.xsd">
+    <description>The coding standard for PHP_CodeSniffer itself.</description>
+    <file>source</file>
+    <exclude-pattern>*source/Application/views/.*$</exclude-pattern>
+    <exclude-pattern>*source/tmp/.*$</exclude-pattern>
+    <arg name="colors"/>
+    <arg value="p"/>
+    <arg value="n"/>
+    <arg name="parallel" value="75" />
+    <arg name="extensions" value="php"/>
+    <rule ref="PSR12">
+        <exclude name="PSR2.Methods.MethodDeclaration.Underscore"/>
+    </rule>
+</ruleset>
--- a/source/Application/Component/UserComponent.php
+++ b/source/Application/Component/UserComponent.php
@@ -257,7 +257,7 @@ class UserComponent extends \OxidEsales\Eshop\Core\Controller\BaseController
      * Executes oxcmp_user::login() method. After loggin user will not be
      * redirected to user or payment screens.
      */
-    public function login_noredirect() //phpcs:ignore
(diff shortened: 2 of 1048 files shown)


phpcs.xml.dist
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0"?>
+<ruleset xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="PHP_CodeSniffer" xsi:noNamespaceSchemaLocation="phpcs.xsd">
+    <description>The coding standard for PHP_CodeSniffer itself.</description>
+    <file>source</file>
+    <arg name="colors"/>
+    <arg value="p"/>
+    <arg name="parallel" value="75" />
+    <arg name="extensions" value="php"/>
+    <rule ref="PSR2"/>
```
alfredbez: I think we use PSR12 in the wrapper:
```suggestion
    <rule ref="PSR12"/>
```

Take a look into: https://github.com/OXID-eSales/coding_standards_wrapper/blob/v3.0.0/bin/oe-eshop-phpcs_with_thresholds#L62

flow-control: :+1: Nice catch, but i needed to change more. I cleaned up and force pushed :wink:

flow-control: FYI: having underscores in method names was a warning in PSR2, but it is an error in PSR12, so i had to exclude this for now from the ruleset, and let the PHP CBF cleanup again :wink:

source/Application/Component/UserComponent.php
```diff
@@ -260,7 +260,7 @@ protected function _afterLogin($oUser)
      * Executes oxcmp_user::login() method. After loggin user will not be
      * redirected to user or payment screens.
```
alfredbez: I think it's cleaner to ignore these errors in the docblock, like:
```suggestion
     * redirected to user or payment screens.
     *
     * @phpcs:disable PSR1.Methods.CamelCapsMethodName.NotCamelCaps
```
You can find the violated rule by adding the `-s` parameter to `phpcs`.

The benefit is, that it does not break static code analysis.

flow-control: Hey @alfredbez,

thanks. Sadly phpcs is not working with PHPDoc annotations. So this change would disable this rule from that line onward till the end of the file, which is not what we want. But I added the rule after the `phpcs:ignore` comment, see commit https://github.com/OXID-eSales/oxideshop_ce/pull/769/commits/1b9e4a2eb1f0b4711e09c8655990ebbfcdee1cad

Could you further explain what you mean by

> does not break static code analysis

As I did not see any difference in the report from PHPStan with or without that `phpcs:ignore` comment on the `function` line.

/Flo

alfredbez: You're right, I just saw this in https://github.com/squizlabs/PHP_CodeSniffer/issues/2367.

The comment shouldn't be a problem for static code analyzers either, just tested it here: https://psalm.dev/r/1efb47d890

It is only a problem if you make the comment on a seperate line between the docblock and the function like here: 
https://psalm.dev/r/1c5ba583c6

flow-control:
Rebased on latest `master`

flow-control:
- rebased on `b-6.2.x` branch and target the PR against it
- as discussed in `OXDEV-3281` I will deprecate all `_` methods in this PR also
- and will create new methods without `_`

flow-control:
As discussed in last knowledge exchange, I will merge this to `b-6.2.x` and then to `master`

PR #771, opened 2020-03-03
Update field to improve performance

AlfonsMartin:
I updated the setter within Fields.php

We didn't change to overall logic. But we gained some minor performance improvement while getting a big readability improvement.

@flow-control was kind enough to make some perfomance checks

**Before update**

`
benchValueFromField.......I0 [μ Mo]/r: 8.280 8.280 (μs) [μSD μRSD]/r: 0.000μs 0.00%
 benchRawValueFromField.....I0 [μ Mo]/r: 1.955 1.955 (μs) [μSD μRSD]/r: 0.000μs 0.00%
 benchValueFromRawField.....I0 [μ Mo]/r: 2.026 2.026 (μs) [μSD μRSD]/r: 0.000μs 0.00%
 benchRawValueFromRawField..I0 [μ Mo]/r: 2.564 2.564 (μs) [μSD μRSD]/r: 0.000μs 0.00%
`

**After changes**

`
benchValueFromField.......I0 [μ Mo]/r: 7.714 7.714 (μs) [μSD μRSD]/r: 0.000μs 0.00%
benchRawValueFromField.....I0 [μ Mo]/r: 1.961 1.961 (μs) [μSD μRSD]/r: 0.000μs 0.00%
benchValueFromRawField.....I0 [μ Mo]/r: 2.065 2.065 (μs) [μSD μRSD]/r: 0.000μs 0.00%
benchRawValueFromRawField..I0 [μ Mo]/r: 2.285 2.285 (μs) [μSD μRSD]/r: 0.000μs 0.00%
`

This seems not much, but these functions are called a lot. 

image

### DIFF
--- a/source/Core/Field.php
+++ b/source/Core/Field.php
@@ -8,6 +8,7 @@
 namespace OxidEsales\EshopCommunity\Core;
 
 use OxidEsales\Eshop\Core\Str;
+use function is_string;
 
 /**
  * Database field description object.
@@ -60,14 +61,11 @@ class Field // extends \OxidEsales\Eshop\Core\Base
     {
         // duplicate content here is needed for performance.
         // as this function is called *many* (a lot) times, it is crucial to be fast here!
-        switch ($type) {
-            case self::T_TEXT:
-            default:
-                $this->rawValue = $value;
-                break;
-            case self::T_RAW:
-                $this->value = $value;
-                break;
+        $this->rawValue = $value;
+        if ($type == self::T_RAW) {
+            $this->value = $value;
+        } else {
+            $this->value = Str::getStr()->htmlspecialchars($value);
         }
     }
 
@@ -80,16 +78,7 @@ class Field // extends \OxidEsales\Eshop\Core\Base
      */
     public function __isset($name)
     {
-        switch ($name) {
-            case 'rawValue':
-                return ($this->rawValue !== null);
-                break;
-            case 'value':
-                return ($this->value !== null);
-                break;
-            //return true;
-        }
-        return false;
+        return ($this->{$name} !== null);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


source/Core/Field.php
```diff
@@ -80,16 +77,7 @@ public function __construct($value = null, $type = self::T_TEXT)
...
-            case 'rawValue':
-                return ($this->rawValue !== null);
-                break;
-            case 'value':
-                return ($this->value !== null);
-                break;
-            //return true;
-        }
-        return false;
+        return ($this->{$name} !== null);
```
alfredbez: Maybe we can get rid of the parentheses here?
```suggestion
        return $this->{$name} !== null;
```

flow-control:
I like it @AlfonsMartin

- it makes the code readable
- and more expressive
- does not change behaviour
- is proven to be faster in 1 of 4 cases
- and performs the same in the other 3 of 4 cases

LGTM!

source/Core/Field.php
```diff
@@ -60,14 +62,9 @@ public function __construct($value = null, $type = self::T_TEXT)
...
         // as this function is called *many* (a lot) times, it is crucial to be fast here!
-        switch ($type) {
-            case self::T_TEXT:
-            default:
-                $this->rawValue = $value;
-                break;
-            case self::T_RAW:
-                $this->value = $value;
-                break;
+        $this->rawValue = $value;
```
Sieg, line 65: Looks like this point should go to "else", no?

AlfonsMartin: I think it's not needed, because of the behavior of the `__get` function.

Sieg: In case of get('value'), the htmlspecialchars part of getter will be additionally called if we constructed as RAW. Have i missed something? Not sure if its good or bad, but the flow looks not as it was before anymore :)

AlfonsMartin: The htmlspecialchars will be called everytime get('value') is called anyway.
We did change the flow of setting just a little bit to save some time, but it doesn't matter because the _get('value') isn't affected at all.

flow-control: @Sieg to me it looks like internal flow changed a bit, but outside behavior stayed the same

Sieg: Yah, looks like i have missed that those are magic methods and are not called when we already have those values filled. But here is another question - We have _initValue method, which basically is for initializing the value, maybe it would be clever to use that one here?

PR #781, opened 2020-03-06
Alternative to DbMetaDataHandler::fieldExists() or tableExists() a other concept

TumTum:
Sometimes the event 'onActivate' is used, when a module is activated. To adjust the database tables. For this purpose the DbMetaDataHandler class was used to adjust the tables.

My idea would be to use a library. That does a check of the database structure and automatically adds the columns or creates the whole table if it doesn't exist.

The developer writes exactly what he needs and the rest does the framework.

Example:

> module/my/metadata.php

```php
'events' => [
      'onActivate'   => 'my\\ModuleInit::onActivate',
      'onDeactivate' => 'my\\ModuleInit::onDeactivate'
  ],
```

> module/my/myModuleInit.php

```php
namespace my;

use OxidEsales\EshopCommunity\Internal\Framework\Module\Setup\Template\ModuleEventsInterface;
use tm\oxid\SchemaExpander\DesireExpander;

class ModuleInit implements ModuleEventsInterface
{
    /**
     * What is new: the onActivate function was given the library as parameter.
     * Which takes care that the table looks like it was described.
     *
     * @param DesireExpander $desireExpander
     */
    public function onActivate(DesireExpander $desireExpander)
    {
        // Simple create new table
        $desireExpander
            ->table('tm_example')
                ->addFieldOxid()
                ->addFieldOxactive()
                ->addFieldOxactiveFrom()
                ->addFieldOxactiveTo()
                ->addField('OXHASH', "char(32) COLLATE latin1_general_ci NOT NULL DEFAULT '' COMMENT 'Hash'")
                ->addField('OXTIME', "int(11) NOT NULL COMMENT 'Validation time'")
                ->addFieldOxtimestamp()
                ->setPrimaryKey('OXID');

        // Extent a oxarticles table
        $desireExpander
            ->table('oxarticles')
                ->addField('MYCOLUMN', "char(32) NOT NULL DEFAULT 'Wowo' COMMENT 'Extent only one Column'")
                ->after('oxlang');
    }

    public static function onDeactivate()
    {
    }
}
```

If you like that idea, I can write the appropriate tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "symfony/console": "^v3.4.15",
         "webmozart/path-util": "^2.3",
         "symfony/finder": "^3.4",
-        "symfony/filesystem": "^3.4"
+        "symfony/filesystem": "^3.4",
+        "tumtum/oxid-schema-expander": "^1.0"
     },
     "require-dev": {
         "oxid-esales/flow-theme": "dev-master",
--- a/source/Internal/Framework/Module/Setup/DataObject/ParameterDatabaseSchemaExpander.php
+++ b/source/Internal/Framework/Module/Setup/DataObject/ParameterDatabaseSchemaExpander.php
@@ -1,0 +1,50 @@
+<?php
+/**
+ * Copyright © OXID eSales AG. All rights reserved.
+ * See LICENSE file for license details.
+ */
+
+namespace OxidEsales\EshopCommunity\Internal\Framework\Module\Setup\DataObject;
+
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use OxidEsales\EshopCommunity\Internal\Framework\Module\Setup\Event\FinalizingModuleActivationEvent;
+use tm\oxid\SchemaExpander\DesireExpander;
+
+/**
+ * Class ParameterDatabaseSchemaExpander
+ * @package OxidEsales\EshopCommunity\Internal\Framework\Module\Setup\DataObject
+ */
+class ParameterDatabaseSchemaExpander implements ParameterInterface, EventSubscriberInterface
+{
+    /**
+     * @var DesireExpander;
+     */
+    private $desireExpander;
+
+    /**
+     * Create new Paramerter
+     *
+     * @return DesireExpander
+     */
+    public function getParameter()
+    {
+        return $this->desireExpander = new DesireExpander();
+    }
+
+    /**
+     * Change Database Schema
+     */
+    public function commitDatabaseSchema()
+    {
+        if ($this->desireExpander) {
+            $this->desireExpander->execute();
+        }
+    }
+
+    public static function getSubscribedEvents()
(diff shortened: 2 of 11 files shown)


Sioweb:
Using a shema builder, is a great idea.

Oxid use doctrine to do migrations, maybe we should also think about using its querybuilder, or entities in some future. It has a lot of modular features and a big community.

alfredbez:
Not directly related, but we still need some way to change the database structure for already active modules. Maybe the `onActivate` event should not contain the schema changes itself, but only trigger them. Additionally, we need a way to trigger them for already active modules, e.g. with the migration-command.

---

Just an idea: Put this migration / schema-definition in a file (e.g. `schema.php`, `migration.php`, `database.php`, etc.) and make it a default `onActivate` event.

Example:
1. We have a module with a `schema.php` (contains DesireExpander / migrations)
2. If you activate the module, the shop detects the `schema.php` during the shop-default `onActivate` event (without the need to the `'onActivate'   => 'my\\ModuleInit::onActivate'` entry in the `metadata.php`)
3. Something in the module changed and we need an additional field or table
4. You adapt the `schema.php`
5. During the deployment, the command `oe-eshop-db_migrate` will execute the `schema.php` for all active modules
6. You have the changed schema, no need to deactivate and re-activate the module

Sioweb:
Thats also possible with Doctrine, u just add the wantet shema and docrine will check em all and build it with - for example - the comand line: `vendor/bin/console doctrine:schema:update -f`

IgorIe:
You don't need to use onActivation at all, there is FinalizingModuleActivationEvent, you can register a subscriber in your module and do whatever you want on this event, use any third party libraries. You already have access to the doctrine query builder via service QueryBuilderFactoryInterface, just inject it in your subscriber and use it.

We have also plans to implement migrations (maybe doctrine migrations) for modules, what should fix the problem.

TumTum:
It doesn't matter if you use common third party libraries to handle the DB communication. However, standardized columns such as OXID, OXACTIVEFROM, OXTIMESTAMP, and so on should be easy to call using a function. Not that everyone tinkers like his own SQL.

@alfredbez in your method. I've made sure the module can't work unless the schematic is changed. With the automatic activation, the DB is finished and the module is active. You don't have to go back in through the CLI and process it again.

@Sioweb ContainerFactory::getInstance()->getContainer()->get(QueryBuilderFactoryInterface::class)->create() Then you already have your QueryBuilder.

Sieg:
Hello guys. Thanks for your input in improving the installation process of the modules.

For now, as Igor already mentioned, there are a couple of ways how you can do suggested things for your projects - please use them. Later on, with improving the migrations system to support module migrations, everything suggested here, will be done much simpler as well, and this was mentioned by Igor too.

TumTum:
Okay. I will wait to see what's going to happen.

PR #782, opened 2020-03-24
Pass optional cookie parameters to setSessionCookie

cpjolly:
Adds the optional parameters used by Registry::getUtilsServer()->setOxCookie(). For example, in some applications it is necessary that the expire time is set to a value and not the default of 0, so that the session can be controlled to live longer than the current browser session.

Adds a standard function comment header.


Sieg:
Hey @cpjolly! But thats private method, and there are no usages with your suggested params. Whats the point? I am probably missing something.

cpjolly:
Hi @Sieg 
- The private function setSessionCookie is new
- The code in setSessionCookie used to be in the public function initNewSession
- I override initNewSession and I must pass extra parameters to setOxCookie
- I must control the cookie lifetime because this is what is needed by my customer
- With default of 0 passed to setOxCookie, the cookie lifetime is alway the browser session lifetime
- With this PR the default is unchanged but all the other setSessionCookie parameters are accessible
- The parameters on setSessionCookie need to be accessible

I hope this explains the situation ?

Sieg:
Hey @cpjolly! I am still missing the main point - even if you have owerwritten the initNewSession in your module, it does not answer the question how this private method change will help you, as you cannot call it from your code. Please double-check if you have pushed everything you wanted :)

Sieg:
Thanks for pointing out the issue!

flow-control:
Hey @cpjolly,
I double checked this and wanted to ask if you can solve this be overloading `OxidEsales\EshopCommunity\Core\UtilsServer::setOxCookie` directly?

```php
class MyClass extends MyClass_parent
{
    public function setOxCookie($sName, $sValue = "", $iExpire = 0, $sPath = '/', $sDomain = null, $blToSession = true, $blSecure = false, $blHttpOnly = true) {
        return parent::setOxCookie($sName, $sValue, time()+86400, $sPath, $sDomain, $blToSession, $blSecure, $blHttpOnly);
    }
}
```

This is a very simple example, but should be an option. Ping me if you see problems with this proposed solution in your specific case and we can have a deeper look into it.

Kind regards
Florian

cpjolly:
Hi @Sieg - my mistake, you're correct. The proposal doesn't work unless the function setSessionCookie is marked protected. Actually, isn't the simplest solution of all to mark setSessionCookie as protected without any other code changes, then it can be overridden as needed ?

Hi @flow-control - it could also be done that way. In my particular case I will need to check the value of $sName to only override the certain Cookies for the session.

source/Core/Session.php
```diff
@@ -1121,13 +1121,26 @@ protected function sidToUrlEvent()
...
+     * @param int    $expire      Expire time
+     * @param string $path        The path on the server in which the cookie will be available on
+     * @param string $domain      The domain that the cookie is available.
+     * @param bool   $toSession   If true, records cookie information to session
+     * @param bool   $secure      If true, transfer cookie only via SSL
+     * @param bool   $httpOnly    If true, only accessible via HTTP
+     *
+     * @return void
+     */
+    private function setSessionCookie($sessionId, $expire = 0, $path = '/', $domain = null, $toSession = true, $secure = false, $httpOnly = true): void
```
flow-control:
```suggestion
    /**
     * Set the session cookie
     *
     * @param string $sessionId   Session cookie value
     *
     * @return void
     */
    protected function setSessionCookie($sessionId): void
```

source/Core/Session.php
```diff
@@ -1121,13 +1121,26 @@ protected function sidToUrlEvent()
...
     {
         if ($this->_getSessionUseCookies()) {
             if (!$this->_allowSessionStart()) {
                 Registry::getUtilsServer()->setOxCookie($this->getName(), null);
             } else {
-                Registry::getUtilsServer()->setOxCookie($this->getName(), $sessionId);
+                Registry::getUtilsServer()->setOxCookie($this->getName(), $sessionId, $expire, $path, $domain, $toSession, $secure, $httpOnly);
```
flow-control:
```suggestion
                Registry::getUtilsServer()->setOxCookie($this->getName(), $sessionId);
```

cpjolly:
Hi @flow-control - thanks for the feedback.

I'm happy to rebase to b-6.2.x and to update the pull request etc as you suggest. Could you please point me at some GitHub or other documentation that describes the exact steps I need to take to rebase my fork and then modify & resubmit the PR so that it updates this PR and doesn't create an unnecessary new one ?

flow-control:
Hey @cpjolly,

I see you have not created a new branch to work on this, but worked in your forks `master` branch. That makes it complicated :wink: and I don't know if there is any good/easy solution for this.

Maybe its best if you do the following:

- checkout `b-6.2.x` branch from your fork
- create a new branch from that in your fork
- cherry pick this one commit into your branch
- and open a new pull request

I don't know another way how to recover this from here. If you had created another branch from `master` you could have used `git rebase --onto b-6.2.x master your-branch-name` (this may also work with commit hashes, but I have never done it bevor)

Hope I could help

/Florian

cpjolly:
@Sieg , @flow-control  - I wasn't able to find a way to update this pull request with the requested changes, so I have issued a new pull request #785. Note that the new pull request does not yet include the requested entry in Changelog.md. Once you're happy with the code in #785, I will write the changelog text.

PR #784, opened 2020-03-30
OXDEV-3281 deprecate all underscore methods

flow-control:
Hello there :open_hands:,

this pull request will deprecate most methods that start with an `_` as stated in OXDEV-3281. In `master` these methods will be renamed to be without the prefixed `_`, to be released with the next major.

/Flo


flow-control:
rebased on latest `b-6.2.x`

PR #785, opened 2020-03-31
Make setSessionCookie protected, correct _sessionStart regression

cpjolly:
This PR implements the requested changes and replaces PR #782. It also corrects a regression.

1. Make the setSessionCookie function protected, so that modules that extend Session can modify the parameters passed to setOxCookie, for example to change the session lifetime.

2. Add a standard comment header to setSessionCookie.

3. Correct a regression in _sessionStart between master and b-6.2.x, where some redundant code that is removed in master has reappeared.
See lines 313-318 in
https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/Session.php
See lines 312-329 in
https://github.com/OXID-eSales/oxideshop_ce/blob/b-6.2.x/source/Core/Session.php

### DIFF
--- a/source/Core/Session.php
+++ b/source/Core/Session.php
@@ -309,23 +309,11 @@ class Session extends \OxidEsales\Eshop\Core\Base
      */
     protected function _sessionStart() // phpcs:ignore PSR2.Methods.MethodDeclaration.Underscore
     {
-        if (!headers_sent() && (PHP_SESSION_NONE === session_status())) {
-            if ($this->needToSetHeaders()) {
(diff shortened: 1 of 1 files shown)


Sieg:
Hey @cpjolly! Thanks for your try to rebase it! :) I have fixed the problem with _sessionStart code, and merged at 1a22016bc7bfc427033349c020bd9518a2fa5fd6. Have a nice day!

cpjolly:
Many thanks @Sieg and @flow-control .

PR #787, opened 2020-04-06
More config-records are selected than it should

JaroslavHerber:
The "%" in the LIKE-statement selects more config-records than it should.
For example, if you have 2 themes "mytheme" and "mytheme-mobile" and lets say "mytheme" is the active one, this query selects all configs from both themes, because LIKE "mytheme%" also matches "mytheme-mobile".
Whats the purpose of this "%" anyway?


source/Core/Config.php
```diff
@@ -566,7 +566,7 @@ protected function _loadVarsFromDb($shopId, $onlyVars = null, $module = '') // p
 
         if ($module) {
             $select .= " oxmodule LIKE :oxmodule";
```
flow-control, line 568:
```suggestion
            $select .= " oxmodule = :oxmodule";
```

flow-control:
Hey @JaroslavHerber,
thanks for taking the time to make this PR. We will discuss this with the team, why we have this `LIKE` condition there. In the mean time, could you cleanup this a bit more?
- as we do not use the`%` anymore, it should not be a `LIKE` but an `=`
- considering this, we might not need that `if else`-block at all anymore and could just add that to the `where`-condition in line https://github.com/OXID-eSales/oxideshop_ce/blob/0863e2050820ab8d7fdd97f37c561cbf6bf40721/source/Core/Config.php#L565
- and we could also move the binding of the `$module` parameter up to line https://github.com/OXID-eSales/oxideshop_ce/blob/0863e2050820ab8d7fdd97f37c561cbf6bf40721/source/Core/Config.php#L559

What do you think about this?

/Florian

JaroslavHerber:
Hey @flow-control,

thank you for your comment. Meanwhile I figured out, why the "%" is there. It's because of this const:
`const OXMODULE_MODULE_PREFIX = 'module:';`

.. which is used in init() to select and load all module-configs at once:
`// loading modules config`
`$this->_loadVarsFromDb($shopID, null, Config::OXMODULE_MODULE_PREFIX);`

I'll reconsider a better solution...

Best regards, Jaro

flow-control:
Hey @JaroslavHerber,

we will run this branch through our internal CI and merge if everything is green.
FYI: our internal CI system also runs unit/integration tests with PE and EE editions and several modules installed, as well as Codeception tests.

/Flo

flow-control:
Hey @JaroslavHerber,

just an update for you: I rebased your pull request on latest `master` and fixed some minor coding style issues. Also this should fix some tricky issues in travis. Lets hope for the best :crossed_fingers: 

/Flo

flow-control:
Hello @JaroslavHerber,

thank you very much, internal CI is also :green_heart:, just merged everything.

/Flo

JaroslavHerber:
Hi @flow-control ,

thats awesome. Thank you very much :)

Best regards,
Jaro

PR #788, opened 2020-04-12
Update .gitignore

8i11y:
The ignored directories where fully ignored (within the exceptions) before.
Now the exceptions will not be ignored anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,7 @@
 !/source/tmp/.htaccess
 
 # Composer
-#/vendor/
+/vendor/
 /composer.lock
 
 # Testing library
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -2,24 +2,28 @@
...
 
 # Temporary files
-/source/tmp/
+!/source/tmp/
+/source/tmp/*
 !/source/tmp/.htaccess
 
 # Composer
-/vendor/
+#/vendor/
```
flow-control, line 26: Just out of curiosity: Why would you not ignore the `vendor` directory? As not only it's contents, but the directors itself is fully managed by composer.

8i11y: oh yeah sry, that's a mistake I added the vendor directory to my git repo.... so you can decomment this so that it is active
^^

flow-control:
Thanks @8i11y

PR #789, opened 2020-04-14
Create .gitkeep

8i11y:
.gitkeep for the case the admin don't install the demo-articles


Sieg:
Thanks @8i11y for the Pull request.

PR #790, opened 2020-04-14
Update .gitignore

8i11y:
Deleted all the entries I don't have on my oxid eshop 6.2 installation (maybe ghostfiles / -directories from older versions !? )

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,24 +26,6 @@
 /vendor/
 /composer.lock
 
-# Testing library
-/test_config.yml
-/source/Services/
-
-# Editions
-/source/Edition/
-
-# IDE
-/.idea/
-/.phpstorm-oxid.meta.php
-/.ide-helper.php
-
-# OS
-.DS_Store
-
-# DI cache
-/source/Internal/Container/container_cache.php
 
 # DI for modules and packages
 /var/generated_services.yaml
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -26,24 +26,6 @@
...
 /composer.lock
 
-# Testing library
-/test_config.yml
-/source/Services/
-
-# Editions
-/source/Edition/
-
-# IDE
```
Sieg, line 36: IDE section is used, it ignores everything IDE is generating

.gitignore
```diff
@@ -26,24 +26,6 @@
...
-/.idea/
-/.phpstorm-oxid.meta.php
-/.ide-helper.php
-
-# OS
```
Sieg, line 41: Mac os is generating those files, should be ignored as well

8i11y: ok - so these are all things which are generated in some cases
Didn't know if these are already needed or remained from older versions of oxid ^^
so I wrote it here as a pull request as a hint and you have a look at it ;)

Sieg:
Hey @8i11y, thanks for triggering this and your suggested changes. I have reverted the part that is still used and merged the PR. Have a nice day!

8i11y:
thx Sieg, have also a nice day :)

PR #792, opened 2020-05-04
added deltime for variants

ivoba:
This PR adds min & max deliverytimes inputs and units select to variants admin templates.
So you can specify deliverytimes per variant.
This should work without backend changes and should also work in themes.

Further it fixes an indent in article_stock.tpl


flow-control:
Hey @ivoba,

thanks for taking your time to create this pull request. We talked about this and came to the conclusion to not merge it. Let me tell you why:
- one can already add this information by
    - selecting the variant in the dropdown on the **Main** tab and then proceed to the **Stock** tab
    - hit the **Edit** button on the **Variant** tab and proceed to the **Stock** tab
- this overview is meant for easy access to a minimal set of data about variants
- one can still add this functionality - if needed - by extending the smarty block `admin_article_variant_listitem`

Nevertheless we would merge the indent fixes in `article_stock.tpl`, feel free to open another PR for this.

I hope I could anyway.

Kind regards
Florian

ivoba:
@flow-control Ok agree :)
I wasn't aware of the "Gehe zu" Dropdown, which indeed makes this behaviour possible.
Thanks for pointing this out.

PR #793, opened 2020-05-04
SimpleXml / objectToXml() does not escape special characters like & correctly

dx-bhesse:
addChild() in SimpleXml.php does not escape some special chars and objectToXml() will generate broken XML.
If for example a value with a '&'-character would appear in the value of an object it would not get escaped to &amp; and the resulting XML will be invalid (->asXml() will actually also generate a php warning)

More information about this issue: https://stackoverflow.com/a/555039 / https://bugs.php.net/bug.php?id=45253


Sieg:
Hey @dx-bhesse, thanks for the fix!

PR #794, opened 2020-05-05
turning off autocomplete for smtp fields

keywan-ghadami-oxid:
because this prevents the browser to fill in wrong data in empty fields and that prevents antecedently wrong smtp configuration or personal passwords to be stored in that field.

That fields are used to be empty in environments where a local mail forwarder is installed.


flow-control:
Hey @keywan-ghadami-oxid,
thank you very much for this enhancement :+1:

PR #795, opened 2020-05-12
getRenderer() shouldn't be private

d3datadev:
Every module author should be able to use this method, like in sendOrderEmailToUser()

### DIFF
--- a/source/Core/Email.php
+++ b/source/Core/Email.php
@@ -347,7 +347,7 @@ class Email extends PHPMailer
      *
      * @return TemplateRendererInterface
      */
-    private function getRenderer()
+    protected function getRenderer()
     {
         $bridge = $this->getContainer()->get(TemplateRendererBridgeInterface::class);
         $bridge->setEngine($this->_getSmarty());


flow-control:
Hey @d3datadev :vulcan_salute:,

thanks for taking the time to prepare this pull request. I followed the discussion in `OXID Devs` Slack and had a look into the proposed changes. I understand, that you can not call the `getRenderer` method from modules (as they are extending the Email class and `getRenderer` is `private`).

But we can not accept this pull request against `master`, let me tell you why:
- the `private function getRenderer()` method was introduced so we can upgrade the `public function send*()` methods to use the `TemplateRendererBridge` from internal namespace and still be backward compatible in case someone extended the `Email::_getSmarty` method (which is also deprecated btw.)
- current plans suggest we remove the `getRenderer` in `master` sooner or later
- the preferred way of fetching the renderer from OXID is through the DI container, you can see a nice example at

https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/Email.php#L917

The only idea we came up with was accepting this PR against `b-6.2.x` as `final protected` method, but as we plan on removing this, we would have to instantly deprecate it. I guess this will not bring any value to anyone :wink: 

I hope I could help anyway.

/Flo

PR #709, opened 2019-06-05
trim default value to detect empty string in Database class

alfredbez:
If you try to save a field without a value the `default_value` will be inserted instead. If the default value is an empty string you will end up filling `''` into the row.

----

Take a look at this:
```
MariaDB [shop]> DESC oxmanufacturers;
+---------------+--------------+------+-----+---------------------+-------------------------------+
| Field         | Type         | Null | Key | Default             | Extra                         |
+---------------+--------------+------+-----+---------------------+-------------------------------+
| OXID          | char(32)     | NO   | PRI | NULL                |                               |
| OXMAPID       | int(11)      | NO   | MUL | NULL                | auto_increment                |
| OXSHOPID      | int(11)      | NO   | MUL | 1                   |                               |
| OXACTIVE      | tinyint(1)   | NO   | MUL | 1                   |                               |
| OXICON        | varchar(128) | NO   |     |                     |                               |
| OXTITLE       | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC   | varchar(255) | NO   |     |                     |                               |
| OXTITLE_1     | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC_1 | varchar(255) | NO   |     |                     |                               |
| OXTITLE_2     | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC_2 | varchar(255) | NO   |     |                     |                               |
| OXTITLE_3     | varchar(255) | NO   |     |                     |                               |
| OXTITLE_4     | varchar(255) | NO   |     |                     |                               |
| OXTITLE_5     | varchar(255) | NO   |     |                     |                               |
| OXTITLE_6     | varchar(255) | NO   |     |                     |                               |
| OXTITLE_7     | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC_3 | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC_4 | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC_5 | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC_6 | varchar(255) | NO   |     |                     |                               |
| OXSHORTDESC_7 | varchar(255) | NO   |     |                     |                               |
| OXSHOWSUFFIX  | tinyint(1)   | NO   |     | 1                   |                               |
| OXTIMESTAMP   | timestamp    | NO   |     | current_timestamp() | on update current_timestamp() |
+---------------+--------------+------+-----+---------------------+-------------------------------+
[...]

PR #796, opened 2020-05-13
trim default value in Database class also for non empty strings (MariaDb)

dx-bhesse:
Related to: https://github.com/OXID-eSales/oxideshop_ce/pull/709

If a table has a `VARCHAR`-column with `DEFAULT 'EUR'`, `$item->default_value` gets set to `'EUR'` instead of `EUR` on some newer MariaDB versions
=> The quoting issue not only appears for DEFAULT ''

This pull-request suggests to simply always trim quotes from default-values.
Drawback: Default-Values that actually have quotes at the beginning or end would lose the quotes (never noticed such default-values in real projects yet). 

If quotation marks cannot be trimmed in general, it would be necessary to explicitly check whether MariaDB is being used (i.e. the DB variables `version` and `version_comments` would have to be queried via an extra SQL query)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [unreleased]
+
+### Fixed
+- MariaDB column default values must be trimmed [PR-796](https://github.com/OXID-eSales/oxideshop_ce/pull/796)
+
 ## [6.5.5] - 2020-05-05
 
 ### Deprecated
--- a/source/Core/Database/Adapter/Doctrine/Database.php
+++ b/source/Core/Database/Adapter/Doctrine/Database.php
@@ -1086,6 +1086,10 @@ class Database implements DatabaseInterface
             $characterSet = $this->getMetaColumnValueByKey($column, 'CharacterSet');
             $collation = $this->getMetaColumnValueByKey($column, 'Collation');
 
+            if ($default !== null) {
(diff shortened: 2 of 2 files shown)


flow-control:
Hey @dx-bhesse :vulcan_salute:

thanks for taking the time to prepare this PR.
I could not reproduce this behavior with MySQL 5.7, but digging deeper into this I found [this comment](https://bugs.oxid-esales.com/view.php?id=6914#c12768) in the related bug report which states that:

> This behaviour only happens with a MARIA-DB >= 10.2.*:
> If you use an older Doctrine version with a newer MARIA-DB version...

Playing a while, this brought me to the [doctrine schema manager](https://github.com/doctrine/dbal/blob/2.10.x/lib/Doctrine/DBAL/Schema/MySqlSchemaManager.php#L231) which we sadly :disappointed: do not use.

The drawback you mention is not a drawback in this case, as with the current solution we have exactly the same "problem".

LGTM :+1: 

I will run this PR through our internal CI which has a lot more tests, as this change is very central :wink: and get back with the results.

/Flo

Just FYI: this change was introduced in [MariaDB 10.2.7](https://mariadb.com/kb/en/mariadb-1027-release-notes/) as fix to a bug in SQL standards implementation, so basically following SQL standards from there on. You can find more information in the [linked upstream ticket](https://jira.mariadb.org/browse/MDEV-13132)

flow-control:
Hey @dx-bhesse,

I rebase your changes onto `b-6.2.x` branch and we will see what Travis thinks of that :crossed_fingers: 

/Flo

flow-control:
Your PR is merged into `b-6.2.x` and `master`
Thank you very much :+1:

PR #797, opened 2020-05-15
fix possible bug https://bugs.oxid-esales.com/view.php?id=7137

SeifertDaniel:


flow-control:
Hey @SeifertDaniel :vulcan_salute:

could you please give me more information about the filesystem structure of your module and how your `composer.json` looks like, specifically the `source-directory` and `target-directory` properties.

It looks like a small patch, but TBH with you, we consider rejecting this PR because:

- the `metadata.php` file should go to the package root, alongside the `composer.json`
- this should not work at all with the `metadata.php` residing somewhere else :wink:

Reason we think it should be like this is that the `metadata.php` has no business logik, but "just" communicates to the shop itself how it should be interacted with.

Kind regards

Flo

SeifertDaniel:
Hello Florian,

thank you for your comment.

An example of the typical structure (including composer.json) can be found here: https://git.d3data.de/D3Public/AttributeMultiValues
Although this module itself is not yet available for OXID 6.2.x, the folder structure would be identical.

Your statement (the metadata must be in the root directory) is contradicted by the possibility to specify a source directory in the composer.json of the modules, which is used extensively in the shop. See:

- OxidEsales\ComposerPlugin\Installer\Package\ModulePackageInstaller Ln 133 ff.
- OxidEsales\EshopCommunity\Internal\Framework\Modules\Install\DataObject\OxidEShopPackage Ln 106 + 114
- OxidEsales\EshopCommunity\Internal\Framework\Modules\Install\Service\ModuleInstaller Ln 83
- OxidEsales\EshopCommunity\Internal\Framework\Modules\Install\Service\BootstrapModulesInstaller Ln 46

Why should an obvious error in one occurrence suddenly be a desired behavior?

We like to use the possibility of the defined source directory, because it keeps the structure of the module clear. And what the metadata.php has to do with the necessary folder structure is not revealed to us at all. The metadata validator already checks that no business logic is to be contained in it, so it is guaranteed. So, for what comprehensible reason should the metadata be forced into the root folder?

Personally, I find it regrettable that such absurd restrictions impose clearly disadvantageous specifications on the module developers.

For further exchange on this topic I am at your disposal in the well-known OXID Slack Channel.

Best regards,
Daniel.

flow-control:
Hey @SeifertDaniel,

since you mentioned that this worked in `v6.1` I'll reconsider this PR.

Regarding the placement of the `metadata.php` file:
As we see, it is technically possible to place that file somewhere else and make it end up in `source/modules/[vendor-id]/[package-name]/metadata.php` where the shop is looking for it. Although it is technically possible, does not imply it is a good idea to do so. The `metadata.php` follows the same purpose as the `composer.json` / `services.yaml` / `phpstan.neon` / `phpcs.xml` /`...`, it gives meta information about how to handle this package. Also it is not source code - although one might say it is a PHP file. It's about intend, not about file type.

Also I don't see any disadvantageous specifications that imposes absurd restrictions on module developers. This is just following [Convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration) which makes live easier for everyone involved.

Regarding your module: Thanks very much for pointing me to the source code. I assume your intend there is you do not want the `patches` directory end up in OXID's `module` directory. I support this idea :+1: , but there is an alternative approach. I'll ping you in Slack, maybe we can come up with a blog post or start a thread in the forum, so more people will participate :smile:

Hope I could help anyway and I'll talk with the core team.

/Flo

flow-control:
Hey there,

could you rebase your PR on `b-6.2.x` branch?
Thanks in advance

Florian

SeifertDaniel:
Hi Flo,

I can't switch this PR to another branch without any side effects. I'll close this and (re)open it as https://github.com/OXID-eSales/oxideshop_ce/pull/800

I look forward to the optimisation proposals in our package structure.

Best regard.
Daniel

flow-control:
Hi,

I saw your worked directly in the `master` branch. Just my two cents: create your own branch before making changes, that way you are flexible in two ways:
- you can make multiple pull requests to the same target branch without messing things up
- you can rebase :wink: (also rebase onto totally different branch)

You can find a good read about this here: https://codeinthehole.com/tips/pull-requests-and-other-good-practices-for-teams-using-github/#sample-workflow

Thanks for the new PR, we will check it.

/Flo

PR #798, opened 2020-05-16
Added str_starts_with(), str_ends_with(), str_contains()

dx-bhesse:
to StrMb and StrRegular as they are often needed in oxid-modules (often emulated with strpos, ...)

The behaviour is analog to the upcoming php8 functions:
https://wiki.php.net/rfc/add_str_starts_with_and_ends_with_functions
https://wiki.php.net/rfc/str_contains
The tests for those functions are also analog to the tests that can be found in
https://github.com/php/php-src/tree/31fb6a08b3a02e665d0e24d2cbd56d13342423c8/ext/standard/tests/strings

Also added an IStr-Interface to make it easier to keep StrMb and StrRegular in sync

### DIFF
--- a/source/Core/IStr.php
+++ b/source/Core/IStr.php
@@ -1,0 +1,293 @@
+<?php
+
+/**
+ * Copyright © OXID eSales AG. All rights reserved.
+ * See LICENSE file for license details.
+ */
+
+namespace OxidEsales\EshopCommunity\Core;
+
+/**
+ * Factory class responsible for redirecting string handling functions to specific
+ * string handling class. String handler basically is intended for dealing with multibyte string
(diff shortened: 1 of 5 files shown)


flow-control (approved):

IgorIe:
Hi @dx-bhesse , thank you for your pull request, the idea to have this functionality is very good, but we would like to use symfony string component https://symfony.com/doc/current/components/string.html instead, deprecate and completely remove StrMb and StrRegular classes in the next major.

Pros:
- very easy to use and read
- it has more useful features and functions
- OOP way
- no maintenance from our side

dx-bhesse:
Ok, that even better - and i will close the this pull-request then ;D

PR #801, opened 2020-05-27
0007116: Module class extensions cannot be sorted in backend

dx-bhesse:
This is the suggested solution (jquery-ui upgrade + small bugfix in oxmoduleslist.js) from https://bugs.oxid-esales.com/view.php?id=7116


AlfonsMartin:
Thx for making the pull request to fix it and thanks for checking this one, we needed it for 6.2. So I just took it and put it to the next 6.2. patch.

https://github.com/OXID-eSales/oxideshop_ce/commit/7e2394e14c3ced429fdeba19b7dc4f2bf7010a0b

PR #802, opened 2020-05-28
Throw exception if language could not be loaded

alfredbez:
I stumbled upon [this issue](https://gist.github.com/alfredbez/e03e3797d2cf73c40969b8fa452ea76a) a few times and finally took some time to think about a better fix, than just to fail with a weird error message.

Things to cleanup if you like the idea:

- [x] Use a better fitting exception instead of `\Exception`
- [x] Fix the failing tests
- [x] Add a new testcase if needed
- [x] Check if there is a related error in the psalm-baseline and remove it from there (We returned an `int` instead of a `string`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Don't copy module files to the `source/modules` directory
 - Copy module assets to the shop out directory
 - Updated a list of bots (aRobots array in config) [PR-853](https://github.com/OXID-eSales/oxideshop_ce/pull/853)
+- Throw exception in getLanguageAbbr method if no abbreviation is available by specific id [PR-802](https://github.com/OXID-eSales/oxideshop_ce/pull/802)
 
 ### Deprecated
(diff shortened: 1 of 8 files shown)


Sieg:
Hi @alfredbez, we had a short talk about this - you can continue with writing a test for this case and fixing failing ones! Thanks for your interest in this!

alfredbez:
We should probably change the psalm-error-level from 8 to 6 to find problems with wrong types. The problem is that psalm reports more than 1200 errors with that level. We can ignore them for now, but I think we should tackle that somehow.

alfredbez:
I don't have a test yet to check that `No languages available` is in the exception-message, which will happen if `aLanguageParams` is empty. I would expect, that this should be handled by [this unit test](https://github.com/OXID-eSales/oxideshop_ce/blob/71a85e1eb5b76f453d1205fb127b97edb15b64cd/tests/Unit/Core/LangTest.php#L1055-L1069), but somehow `de` & `en` are still available. Any idea why?

Sieg:
Hey @alfredbez, thanks for the improvement. I have rebased on latest master, added language validation in getActLanguageId so frontend would not explode too much on wrong lang paramater values, and merged! Have a nice day.

PR #803, opened 2020-05-28
Fix return value for Article::getCategory method

flow-control:
Hey there :vulcan_salute: 

the return value of the `OxidEsales\EshopCommunity\Application\Model\Article::getCategory` method is not easy to work with and puts a burden on the caller. This PR tries to fix most of it.

#### Problems with the return value
- you might get `null` or `\OxidEsales\Eshop\Application\Model\Category`
- but `\OxidEsales\Eshop\Application\Model\Category` is not guaranteed to be initialized

Therefore the caller musst always check if the returned value is `null` and if not, call the `getId()` method and check that response additionally to make sure.

#### What this PR changes
Return type is still `Category|null` but with this patches applied, you can not get an uninitialized Category object anymore.

##### Additionally fixed with this PR:
- wrong usage of static member `_aCategoryCache` with `$this` instead of `self`

#### Open problems I see that are not fixed in this PR
- you can not rely on the `Category::isLoaded()` method on the caller side, as the `protected _isLoaded` property is only set in the `load` method wich is not used
- loading of the `Category` object and querying the `oxcategories` table does not belong into the `Article` model

/Flo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Ensure \OxidEsales\EshopCommunity\Application\Model\NewsSubscribed::getOptInStatus int result type
 - Change broken "Requirements" links to current shop documentation [PR-877](https://github.com/OXID-eSales/oxideshop_ce/pull/877)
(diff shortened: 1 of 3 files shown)


Sieg:
Hello @flow-control, thanks for the fix.

PR #806, opened 2020-06-09
fix getting raw password

proudcommerce:
`value` gets encoded data eg. `BsNNF&amp;yQ&amp;F3p`, original: `BsNNF&yQ&F3p` --> smtp authentication failed


Sieg:
Hi @proudcommerce, thanks for the fix! Merged.

PR #807, opened 2020-06-11
Add aliases to the module commands

patrick-blom:
### In a Nutshell
This PR adds aliases to the OXID console module commands to increase typing speed during module and project development.

### Motivation
During module and project development modules will often be activated / deactivated or changed so that the new configuration must be applied. inspired by the Symfony command aliases this PR introduces intuitive aliases for the common OXID module commands e.g. "oe:module:activate".

### Usage
Instead of typing the whole command name the aliases can be used. 

**Example:**
`vendor/bin/oe-console oe:module:activate` becomes `vendor/bin/oe-console m:a`

### Available aliases
- `m:a` => `oe:module:activate`
- `m:d` => `oe:module:deactivate`
- `m:ac` =>  `oe:module:apply-configuration`
- `m:ic` => `oe:module:install-configuration`

### Tests
To ensure that the aliases are functional, the given module command tests were extended.

### DIFF
--- a/source/Internal/Framework/Module/Command/ApplyModulesConfigurationCommand.php
+++ b/source/Internal/Framework/Module/Command/ApplyModulesConfigurationCommand.php
@@ -49,7 +49,8 @@ class ApplyModulesConfigurationCommand extends Command
 
     protected function configure()
     {
-        $this->setDescription('Applies configuration for installed modules.');
+        $this->setDescription('Applies configuration for installed modules.')
+             ->setAliases(['m:ac']);
     }
 
     protected function execute(InputInterface $input, OutputInterface $output)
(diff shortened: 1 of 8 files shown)


Sieg:
Hi @patrick-blom! Thanks for your interest in making OXID better!

Regarding this case - its possible to use those commands with short calls (first letter(s) of every group) without making any alliases, like eg.:

```
bin/oe-console o:m:i source/modules/erp # installs the module with oe:module:install-configuration
bin/oe-console o:m:ac erp # activates the module
```

I think its short enough, and would prefer for everything to stay as it is, without introducing some new hardcoded aliases.

Please let me know if i have missed something.

patrick-blom:
Hey @Sieg,
thx for your answer! 

Even for me, thats short enough ;) I would appreciate if that will be documented in a proper way, because this feature is quite new to me and maybe also to other developers.

flow-control:
Hey @patrick-blom,

as we are using the symfony/console component, it is documented there:
https://symfony.com/doc/current/components/console/usage.html#shortcut-syntax

But you are right, we should probably add this link also to the developer documentation.

/Flo

PR #808, opened 2020-06-15
fix for generated_services in gitignore

ivoba:
Its 
`/var/generated/generated_services.yaml`
not
`/var/generated_services.yaml`


Sieg:
Hey @ivoba! Thanks for the fix!

PR #809, opened 2020-06-16
Performance enhancement

mysteriusguy:
- Instead of iterating "if" statements, using fixed value


Sieg:
Hey @mysteriusguy thanks for the improvement.

PR #810, opened 2020-06-16
Catch circular reference for logger in getLogger()

alfredbez:
We saw this error in the `oxideshop.log`:
```
OXID Logger.ERROR: Circular reference detected for service "Psr\Log\LoggerInterface", path: "Psr\Log\LoggerInterface -> Psr\Log\LoggerInterface". ["[object] (Symfony\\Component\\DependencyInjection\\Exception\\ServiceCircularReferenceException(code: 0)
```
We had this once in a project and three other devs from other projects had the same error.

see also: https://gist.github.com/proudcommerce/602ab33c23a1546588266812ed72ac30

Unfortunately, I have no steps to reproduce again

---

I'm glad if someone finds a better solution to this, just wanted to dump this out here, since that's a better place than a random gist.

### DIFF
--- a/source/overridablefunctions.php
+++ b/source/overridablefunctions.php
@@ -5,8 +5,11 @@
  * See LICENSE file for license details.
  */
 
+use Monolog\Logger;
+use Monolog\Handler\StreamHandler;
 use OxidEsales\Eshop\Core\Registry;
 use OxidEsales\Eshop\Core\Request;
+use Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException;
 
 if (!defined('ESHOP_CONFIG_FILE')) {
     define('ESHOP_CONFIG_FILE', 'config.inc.php');
@@ -204,8 +207,22 @@ if (!function_exists('getLogger')) {
      */
     function getLogger()
     {
-        $container = \OxidEsales\EshopCommunity\Internal\Container\ContainerFactory::getInstance()->getContainer();
+        $loggerServiceName = \Psr\Log\LoggerInterface::class;
+        try {
+            $container = \OxidEsales\EshopCommunity\Internal\Container\ContainerFactory::getInstance()->getContainer();
 
-        return $container->get(\Psr\Log\LoggerInterface::class);
+            return $container->get($loggerServiceName);
+        } catch (ServiceCircularReferenceException $exception) {
+            if (strpos($exception->getMessage(), $loggerServiceName) !== false) {
+                $logger = new Logger('OXID Logger');
+                $logger->pushHandler(new StreamHandler(OX_LOG_FILE, Logger::ERROR));
+
+                $logger->critical('Not able to load logger from container');
+
+                return $logger;
+            }
+
+            throw $exception;
+        }
     }
 }


alfredbez:
Feedback from @BattleMoench:

> for me it was the same as Frank Zunder had: i updated  an existing module via composer. (In this case a new Monolog Logger was inside that module) But why i wasn't able to rollback everything is not clear for me...

alfredbez:
I got the same error just now again, I have more details now.

We're using the Metapackage Version 6.2.2. I was working on a feature branch where we added a new module. After switching to the master branch, I can barely do anything, because this error appears even if I try to run `vendor/bin/oe-console oe:module:apply-configuration`.

Here's the full error:
```
[06 Oct 13:12:37.110807 2020] [uncaught error] [type E_ERROR] [file /var/www/oxideshop/vendor/symfony/dependency-injection/Container.php] [line 297] [code ] [message Uncaught Symfony\Component\DependencyInjection\Exception\ServiceCircularReferenceException: Circular reference detected for service "Psr\Log\LoggerInterface", path: "Psr\Log\LoggerInterface -> Psr\Log\LoggerInterface". in /var/www/oxideshop/vendor/symfony/dependency-injection/Container.php:297
Stack trace:
#0 /var/www/oxideshop/source/overridablefunctions.php(209): Symfony\Component\DependencyInjection\Container->get('Psr\\Log\\LoggerI...')
#1 /var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/Registry.php(329): getLogger()
#2 /var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/Module/ModuleChainsGenerator.php(420): OxidEsales\EshopCommunity\Core\Registry::getLogger()
#3 /var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/Module/ModuleChainsGenerator.php(300): OxidEsales\EshopCommunity\Core\Module\ModuleChainsGenerator->onModuleExtensionCreationError('norisk\\nrClickA...')
#4 /var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/Module/ModuleChainsGenerator.php(251):]
```

---

I hope this helps you to verify this error.

SvenBrunk (approved):
Verified working 👍

IgorIe:
@alfredbez thanks for pointing it out here, looks like you have module class extension data in the database for a no longer existent class, a better solution could be find and fix the circular reference if it's possible, if it's not possible - use another logger in the autoloader (ModuleChainsGenerator class), not a service from the container. Is it reproducible with any module on a clean project?

alfredbez:
Hi @IgorIe, I don't have a clean project at the moment, but I guess it should be reproducible by installing a module on a feature branch and then switching to the master branch again, that's how I can reproduce this in my project.

SvenBrunk:
@IgorIe there is no such thing as a clean project. That is why the shop must be resilient to certain kinds of errors and should be prepared for them. Especially everything from the modules or the configurations can't be trusted entirely.

alfredbez:
I think @IgorIe was talking about something like https://github.com/OXID-eSales/docker-eshop-sdk where I can reproduce the issue.

dx-bhesse:
Another workaround could be to let onModuleExtensionCreationError() only remember the broken $moduleClass in some array and to do the actual error-handling/logging outside onModuleExtensionCreationError() later (for example in the php shutdown-handler).
If the logging is not done inside onModuleExtensionCreationError() i think an endless loop/recursion would not happen (which currently triggers the Circular-Reference-Exception)

For testing the error can be forced by adding a non-existing class in oxconfig.aModules as class-extension for Oxid's "Language"-class (or any other class that Config->init() directly or indirectly uses using oxNew())
Just doing nothing but a return; in onModuleExtensionCreationError() lets oxid load without any issues in most of these cases and in the backend "Installed Shop Modules" then suggests to remove the broken module data...

suabo:
This one hit me again after updating from 4.7 to 6.2. I removed all module settings manually with the known sql query to avoid problems with modules while updating. But now I'm stuck in the upgrade process with this error hitting when i try to execute:

`vendor/bin/oe-eshop-db_views_generate`

Like described as final step of db update [here](https://docs.oxid-esales.com/developer/en/6.0/update/eshop_from_53_to_6/database.html).

I also tried `vendor/bin/oe-console oe:module:apply-configuration` with the same result.

If this one is no bug maybe you could lead to a solution in documentation? 
For now the only solution for me is to use the fix provided to move on.

Indianer3c:
Please, Merge Pull-Request. Same Issue when I'm updating OXID eShop Community Edition 6.2.5 to 6.3.1, the try block from commit work's for me, too.

Indianer3c:
> @alfredbez thanks for pointing it out here, looks like you have module class extension data in the database for a no longer existent class, a better solution could be find and fix the circular reference if it's possible, if it's not possible - use another logger in the autoloader (ModuleChainsGenerator class), not a service from the container. Is it reproducible with any module on a clean project?

You can reproducible by creating a dummy module and reference the version 1.0.0 via packagist.org. Then you change a class name in your dummy module and create new version 1.1.0. When you update the your dummy module in a local environment and then you deploy this changes to your staging environment without reactivating your dummy module.

The try block helps to locate the problem which module have to reactivating, because the Logger writes into log file more information and prevented that the online shops offline. The owner can reactivate the module independently without the help of the developer via the admin.

Indianer3c:
> @Indianer3c could you provide that dummy module code, so it will be easier to reproduce?

Hmm, may you can fork the https://github.com/hkreuter/oxsample_module Then add your fork to packagist.org and change in a new branch the class name from FrontendController to ChangeFrontendControllerName https://github.com/hkreuter/oxsample_module/blob/008-relation_model/metadata.php#L31

Sieg:
Hello guys, thanks for the patience! Most likely this will be merged for 6.5.1.

Sieg:
Hi @alfredbez, sorry for the delay. Could you please sign the updated cla?

alfredbez:
done

Sieg:
Hi @alfredbez, @Indianer3c, @dx-bhesse! I have tried to reproduce this issue on b-7.0.x, looks like this part is handled better here. I cannot break the shop with circular references. Could someone double-check if this is still needed for 7.0.x?

michaelkeiluweit:
@Sieg I tried to reproduce it by following the reproduce steps describes in https://bugs.oxid-esales.com/view.php?id=7286, but the circular references error doesn't occur anymore. It seems like an exception is thrown before so it doesn't reach the critical part:
```php
// oxideshop-ce/source/Internal/Transition/Utility/Context.php
private function getConfigParameter($name, $default = null)
{
    $value = Registry::getConfig()->getConfigParam($name, $default);
    DatabaseProvider::getDb()->setFetchMode(PDO::FETCH_ASSOC);
    return $value;
}
```
Reaching that point causes a connection to the database which itself causes a call of the DI container, which then would fail again.

For example, the callstack of the 6.5.3, when there's a problem with the container file container_cache.php (it reads from bottom to top):
```txt
OxidEsales\EshopCommunity\Internal\Transition\Utility\Context->getConfigParameter (/var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Internal/Transition/Utility/Context.php:147)
OxidEsales\EshopCommunity\Internal\Transition\Utility\Context->getLogLevel (/var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Internal/Transition/Utility/Context.php:33)
OxidEsales\EshopCommunity\Internal\Framework\Logger\LoggerServiceFactory->getMonologConfiguration (/var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Internal/Framework/Logger/LoggerServiceFactory.php:66)
OxidEsales\EshopCommunity\Internal\Framework\Logger\LoggerServiceFactory->getMonologLoggerFactory (/var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Internal/Framework/Logger/LoggerServiceFactory.php:53)
OxidEsales\EshopCommunity\Internal\Framework\Logger\LoggerServiceFactory->getLogger (/var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Internal/Framework/Logger/LoggerServiceFactory.php:43)
OxidEsales\EshopCommunity\Core\Exception\ExceptionHandler->handleUncaughtException (/var/www/oxideshop/vendor/oxid-esales/oxideshop-ce/source/Core/Exception/ExceptionHandler.php:131)
```

Sieg:
@michaelkeiluweit thanks for double-checking, so, as i understood, this issue is not an issue for 7.0 currently anymore? Feel free to reopen this PR is there are still cases not fixed.

PR #813, opened 2020-06-22
Dont show shop versions in html while in productive mode

olivereanderson:
Security improvement suggested by security experts: By displaying the shop's version and edition (as a comment) in the HTML an attacker is able to look more specifically for vulnerabilities in the software used.


Sieg:
Hello @olivereanderson, thanks for the pull request! We had a talk about this with the PO. We cannot just completely remove copyright information from the product, so I have removed version information from copyright string at least - this change was merged to our master recently.

PR #814, opened 2020-06-22
Hide shop version in productive mode.

olivereanderson:
Security improvement suggested by security experts: By displaying the shop's version and edition (as a comment) in the HTML an attacker is able to look more specifically for vulnerabilities in the software used.


Sieg:
Hi @olivereanderson, Thanks for the pull request! Sad to say, but this will not go to 6.1.

PR #815, opened 2020-06-23
Change return type in render method to string

alfredbez:
Psalm reported several problems, because the return type of the `render` method in the `BaseController` and in the `FrontendController` differs:

https://github.com/OXID-eSales/oxideshop_ce/blob/b-6.1.x/source/Core/Controller/BaseController.php#L231

https://github.com/OXID-eSales/oxideshop_ce/blob/b-6.1.x/source/Application/Controller/FrontendController.php#L2106

Maybe we should change both docblocks to use `?string` as a return type or change the [default value of `$_sThisTemplate`](https://github.com/OXID-eSales/oxideshop_ce/blob/b-6.1.x/source/Core/Controller/BaseController.php#L58) to an empty string instead of `null`


Sieg:
Hi @alfredbez, thanks for the improvement. Merged.

PR #816, opened 2020-06-23
Replace getClassId with getClassKey

alfredbez:
Looks like this was a mistake, see 371a61c003037c8b7d6f7529cd64a702ecf8870e


Sieg:
Thanks for the fix @alfredbez, merged!

PR #817, opened 2020-06-26
Remove wrong code for getting module URL from admin

SvenBrunk:
Removing code that assumes that the module URL has anything to do with the admin URL. Probably even the rest of the code can also be simplified a lot.
The code is not just unnecessary, it also causes problems when the shop admin runs under a completely different URL (e.g. a subdomain) than the rest of the shop. (Which is often the case for bigger customers).

image

### DIFF
--- a/source/Core/ViewConfig.php
+++ b/source/Core/ViewConfig.php
@@ -1201,16 +1201,7 @@ class ViewConfig extends \OxidEsales\Eshop\Core\Base
         $shopUrl = null;
         if ($this->isAdmin()) {
             if ($c->isSsl()) {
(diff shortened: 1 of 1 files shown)


SvenBrunk:
There MIGHT be really rare cases where the users of the admin backend can't access the shop frontend, but I guess that is the absolute minority of cases

Sieg:
Hi @SvenBrunk, cleaned up more (https://github.com/OXID-eSales/oxideshop_ce/commit/7acdc25193da261de89e87932999a0138dfecc57), as mentioned by you in the PR description. Merged to b-8.0.x. Please take a look if i have not missed anything :)

PR #819, opened 2020-07-08
Fix wrong return type in language class

alfredbez:
There is a special case in the language class, when `translateString` returns an `array` (for `_aSeoReplaceChars`).

A better solution would be to refactor the code so that this always returns a string.


flow-control:
I second that, but thanks anyway for this contribution :rocket: :+1:

alfredbez:
I think that was the fastest merge I've ever seen for this repository, thanks :tada:

PR #520, opened 2016-12-09
log missing translations

alfredbez:
This PR will enable logging for missing translations.

Questions:
* Do we need an additional flag to enable/disable this kind of logging?
* Do we need more data like shop-id, activated theme or modules?
* Is there a better way to log stuff like this?

I found comments like this
```
//deprecated since v5.3 (2016-06-17); Logging mechanism will be changed in 6.0.
```
over and over again, but I don't know if you have already implemented another logging mechanism.

---

sample output:
```
OxidEsales\EshopCommunity\Core\Exception\StandardException (time: 2016-12-09 13:09:20): [0]: ERROR: Translation for MORE_INFO not found! 
 Stack Trace: #0 /var/www/oxideshop/source/Core/UtilsObject.php(273): OxidEsales\EshopCommunity\Core\UtilsObject->_getObject('oxexception', 0, Array)
#1 [internal function]: OxidEsales\EshopCommunity\Core\UtilsObject->oxNew('oxException')
#2 /var/www/oxideshop/source/oxfunctions.php(317): call_user_func_array(Array, Array)
#3 /var/www/oxideshop/source/Core/Smarty/Plugin/function.oxmultilang.php(90): oxNew('oxException')
#4 /var/www/oxideshop/source/tmp/smarty/6ce77b7a9d9444335a4b8f5ea13cf8cb^%%CB^CB6^CB6C1C15%%listitem_infogrid.tpl.php(224): smarty_function_oxmultilang(Array, Object(Smarty))
#5 /var/www/oxideshop/vendor/smarty/smarty/libs/Smarty.class.php(1270): include('/var/www/oxides...')
#6 /var/www/oxideshop/source/Core/ShopControl.php(469): Smarty->fetch('widget/product/...', 'ox|0|0|0|0')
#7 /var/www/oxideshop/source/Core/ShopControl.php(328): OxidEsales\EshopCommunity\Core\ShopControl->_render(Object(OxidEsales\EshopCommunity\Application\Component\Widget\ArticleBox))
#8 /var/www/oxideshop/source/Core/ShopControl.php(260): OxidEsales\EshopCommunity\Core\ShopControl->formOutput(Object(OxidEsales\EshopCommunity\Application\Component\Widget\ArticleBox))
#9 /var/www/oxideshop/source/Core/ShopControl.php(153): OxidEsales\EshopCommunity\Core\ShopControl->_process('oxwarticlebox', NULL, Array, Array)
[...]

PR #821, opened 2020-07-08
Add missing translation

olivereanderson:
The shop requires a translation corresponding to the key 'PAGE_TITLE_START'. This has already been mentioned in a comment from 23 August 2018 here: https://github.com/OXID-eSales/oxideshop_ce/pull/520. 

To see that this is indeed the case you can for instance set a break point in the `getTitle()` method in the `FrontendController` and refresh the shop's start page. 

https://github.com/OXID-eSales/oxideshop_ce/blob/b-6.2.x/source/Application/Controller/FrontendController.php#L1893


flow-control:
Thank you for this improvement :tada: 
I felt so free to add the English translation as well

/Flo

PR #825, opened 2020-07-21
remove usage of variable as it's not needed

mysteriusguy:


Sieg:
Hey @mysteriusguy, thanks for the improvement.

PR #828, opened 2020-08-13
Improve developer project setup workflow

bpesch:
- Automatically creates copy of config.inc if it doesn't exist yet
- Asks for parameters if no value was given instead of throwing an error directly (user friendliness)

### DIFF
--- a/bin/oe-console
+++ b/bin/oe-console
@@ -9,6 +9,7 @@ use Symfony\Component\Console\Application;
 use OxidEsales\EshopCommunity\Internal\Framework\Console\Executor;
 use OxidEsales\Facts\Facts;
 use Symfony\Component\Console\Input\ArgvInput;
+use Symfony\Component\Filesystem\Filesystem;
 use Webmozart\PathUtil\Path;
 use OxidEsales\EshopCommunity\Internal\Container\ContainerBuilderFactory;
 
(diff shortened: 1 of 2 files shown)


AlfonsMartin:
Hi, 

i like the idea of copying the config.inc. 
But for the prompt, we would then like to have also an --no-interaction paramter, used for automatic deployments which this command should be capable of. And if we adding the prompt, an automatic deployment could get stuck, but we want it to fail. 

Also we need to have some tests regarding this. 
I will discuss this in team.

Sieg:
Hey guys, thanks for your patience with this one. We had several discussions internally regarding the "user friendly" and "magic" topics, that are happening without people knowing about, so at the end, the decisions were:
* lets not add this interaction, as it gives no real benefit - after showing the error of parameter missing, you will recall the previous command again and add the missing parameters. There are no parameters required in the middle of the process, everything is configured at the beginning. Adding the interaction means - later supporting something that is not really needed at all.
* regarding making the copy of the dist config file, we decided to not do it right now, as the config file will get some love soon, and this improvement will be there as well, IF we will still need to create it as it is now.

Sieg:
Also looks like the config.inc.php is copied during the first composer update run, thats another point against adding additional checks and copiyng.

PR #829, opened 2020-08-17
Removed leftover from pre-composer times

mysteriusguy:


Sieg:
Hi @mysteriusguy, thanks a lot for cleaning this up! Merged.

PR #830, opened 2020-08-20
IDE auto completion using oxNew

bpesch:
Works for every Jetbrains IDE capable of PHP.
- [x] IntelliJ IDEA Ultimate 2020.2
- [x] PHPStorm 2020.2

### DIFF
--- a/source/oxfunctions.php
+++ b/source/oxfunctions.php
@@ -92,9 +92,9 @@ function cmpart($a, $b)
  * error message.
  *
  * @template T
- * @psalm-param T $className
+ * @psalm-param class-string<T> $className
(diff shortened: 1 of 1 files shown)


flow-control (approved):

bpesch:
Follow up: The current version seems to support strings like `oxArticle` but not strings like `Article:class`

Sieg:
Hi @bpesch, for me it looks like all the autocompletion works fine with our current version (on 2020.2.1 of phpstorm). Could you please describe more, which case are we trying to fix?

image

image

bpesch:
Hi @Sieg, it seems to work with the newest version 🎉

PR #831, opened 2020-08-24
(Security proposal) Password cannot be username

bpesch:
It's a common practice for modern websites, to check if the username **equals or is similar** to the password, and throw an error if so.
I added an example on how this could be implemented.


flow-control:
Hey there 🖖

I like the idea, I'll check tomorrow.

/Flo

source/Core/InputValidator.php
```diff
@@ -174,37 +176,47 @@ public function checkEmail($user, $email)
...
-     * @param User   $user                      Active user.
+     * @param string $email                     Email.
      * @param string $newPassword               New user password.
      * @param string $confirmationPassword      Retyped user password.
      * @param bool   $shouldCheckPasswordLength Option to check password length.
      *
      * @return Exception\StandardException|null
      */
-    public function checkPassword($user, $newPassword, $confirmationPassword, $shouldCheckPasswordLength = false)
+    public function checkPassword($email, string $newPassword, string $confirmationPassword, bool $shouldCheckPasswordLength = false)
```
Sieg, line 186: We cannot just simply change the parameter ;)

bpesch: Why not? `$user` doesn't seem to be used in the original method.  
Bildschirmfoto 2020-08-24 um 16 50 01. 
https://github.com/OXID-eSales/oxideshop_ce/blob/master/source/Core/InputValidator.php#L184

source/Application/Model/User.php
```diff
@@ -1127,7 +1128,7 @@ public function checkValues($sLogin, $sPassword, $sPassword2, $aInvAddress, $aDe
         $oInputValidator->checkEmail($this, $sLogin);
 
         // 3. password
-        $oInputValidator->checkPassword($this, $sPassword, $sPassword2, ((int) Registry::getConfig()->getRequestParameter('option') == 3));
+        $oInputValidator->checkPassword($sLogin, $sPassword, $sPassword2, ((int) Registry::getConfig()->getRequestParameter('option') == 3));
```
flow-control, line 1131: We can not change what we pass to this method as first argument. We clearly do not use this, but others might chain overload the `InputValidtor` class and rely on the first parameter being the `User` model.
```suggestion
        $oInputValidator->checkPassword($this, $sPassword, $sPassword2, ((int) Registry::getConfig()->getRequestParameter('option') == 3));
```

source/Core/InputValidator.php
flow-control, line 186: As we can not change, what we pass, we should also leave the argument name as `$user`

source/Core/InputValidator.php
```diff
@@ -174,37 +176,47 @@ public function checkEmail($user, $email)
      * Checking if user password is fine. In case of error
      * exception is thrown
      *
-     * @param User   $user                      Active user.
+     * @param string $email                     Email.
```
flow-control, line 179: As we can not change, what we pass to this method, argument should stay the same
```suggestion
     * @param User   $user                      Active user.
```

flow-control (changes requested):
Thanks for your pull request, I marked you some changes we need to make prior to be able to merge this. Additionally could you have a look into the tests?
Thanks in advance
/Flo

bpesch:
Hello,
I'm not comfortable with pursuing old standards.
The tests seem pretty outdated aswell.
If you like the idea, you might be interested in implementing it the way you mentioned. 

Best Regards
Ben

Sieg:
Hey, backward compatibility is something that we are trying to save for developers, so updating to a new shop version would be as painless as it could be! The idea presented here, itself, is good! But we need to think about how to introduce this safer. Have a nice day!
I believe when this part will be refactored, it will be much easier to add new rules and conditions for validating your data.

PR #832, opened 2020-08-28
ditching 'FORM_FIELDSET_USER_*'

kermie:
'FORM_FIELDSET_USER_SHIPPING_ADDITIONALINFO2_TOOLTIP'         => '', // this is specifically for DHL
'FORM_FIELDSET_USER_BILLING_ADDITIONALINFO_TOOLTIP'           => '', // this is specifically for DHL

Ditching 'FORM_FIELDSET_USER_*' as empty lang keys don't make sense. Think of translators!


Sieg:
Hey @kermie, but those translations are used in the templates, despite they are empty at the moment, still can be used for showing some tooltip with information for "Additional info" field.

PR #833, opened 2020-09-09
Streamlined lang files for work with translation platforms

kermie:
* eliminated concatenations in key-value pairs so source can be read for translation platforms (or other systems)
* in en/help_lang.php replaced double quotes and escaped single quote in one string
* corrected distances between keys for better readability


kermie:
Commit 0bd89a3:
* doesn't belong to this file
* is paternalism - also to the community
* we do have other processes to achieve quality
* is basically not correct

Sieg:
Hey @kermie, i have rebased only commits without removing the keys as we have discussed earlier, and merged at bf7aa0503e8974c16f07c3842efc37a06e837d4a

PR #835, opened 2020-09-30
fix for removing bonus article from basket bug:0007061

ivoba:
Possible fix for https://bugs.oxid-esales.com/view.php?id=7061

When removing an item the id is buffered and on discount readding its checked if the id was removed before, so the article is not readded.


Sieg:
Hi @ivoba, thank you for the suggested fix. Sadly, this one will fix one, but will introduce other problems. We will need to rework the mechanism how bonus items are added in general. 

One of the possible solutions could be - show bonus items near by the regular ones in the basket, and let customer decide if one wants it or not.

The bug stays open for now.

PR #837, opened 2020-10-06
Optimize opcode usage

bpesch:
Fist of all, if you're not familiar with opcode, [here's](https://en.wikipedia.org/wiki/Opcode) a definition.

**Purpose of this PR:**
Optimizing the performance by invoking the native PHP functions correctly.  
Without using the incorrect invocation, native functions will be less performant because the initialization and parameter passing is split in separate operations.  

I posted a benchmark and references below.  

**Benchmarks:**
````
<?php //demo.php

namespace Foo;

in_array('foo', ['bar']);
````

produces  

```
/Users/Shared/dev/oxideshop_ce(master*) » php -d vld.active=1 -d vld.execute=0 demo.php                                                                                                                                                                                                         administrator@administorsiMac
Finding entry points
Branch analysis from position: 0
1 jumps found. (Code = 62) Position 1 = -2
filename:       /Users/Shared/dev/oxideshop_ce/demo.php
function name:  (null)
number of ops:  7
compiled vars:  none
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   3     0  E >   NOP
   5     1        EXT_STMT
         2        INIT_NS_FCALL_BY_NAME                                    'Foo%5Cin_array'
         3        SEND_VAL_EX                                              'foo'
         4        SEND_VAL_EX                                              <array>
         5        DO_FCALL                                      0
   6     6      > RETURN                                                   1

branch: #  0; line:     3-    6; sop:     0; eop:     6; out0:  -2
path #1: 0,
```

and  

```
<?php  //demo.php

namespace Foo;

\in_array('foo', ['bar']);
```

produces 

```
/Users/Shared/dev/oxideshop_ce(master*) » php -d vld.active=1 -d vld.execute=0 demo.php                                                                                                                                                                                                         administrator@administorsiMac
Finding entry points
Branch analysis from position: 0
1 jumps found. (Code = 62) Position 1 = -2
filename:       /Users/Shared/dev/oxideshop_ce/demo.php
function name:  (null)
number of ops:  5
compiled vars:  none
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   3     0  E >   NOP
   5     1        EXT_STMT
         2        IN_ARRAY                                         ~0      'foo', <array>
         3        FREE                                                     ~0
   6     4      > RETURN                                                   1

branch: #  0; line:     3-    6; sop:     0; eop:     4; out0:  -2
path #1: 0,
```

**References:**
https://www.php.net/manual/de/internals2.opcodes.list.php
https://github.com/php/php-src/blob/f2db305fa4e9bd7d04d567822687ec714aedcdb5/Zend/zend_compile.c#L3872
https://github.com/Roave/FunctionFQNReplacer#rationale
https://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

### DIFF
--- a/source/Application/Component/BasketComponent.php
+++ b/source/Application/Component/BasketComponent.php
@@ -284,7 +284,7 @@ class BasketComponent extends \OxidEsales\Eshop\Core\Controller\BaseController
 
         // special treatment
         // search param
-        $sParam = rawurlencode(Registry::getConfig()->getRequestParameter('searchparam', true));
+        $sParam = \rawurlencode(Registry::getConfig()->getRequestParameter('searchparam', true));
         $sPosition .= $sParam ? 'searchparam=' . $sParam . '&' : '';
 
         // current page number
@@ -313,10 +313,10 @@ class BasketComponent extends \OxidEsales\Eshop\Core\Controller\BaseController
     protected function getPersistedParameters($persistedParameters = null)
     {
         $persistedParameters = ($persistedParameters ?: Registry::getConfig()->getRequestParameter('persparam'));
-        if (!is_array($persistedParameters)) {
+        if (!\is_array($persistedParameters)) {
             return null;
         }
-        return array_filter($persistedParameters) ?: null;
+        return \array_filter($persistedParameters) ?: null;
     }
 
     /**
@@ -366,7 +366,7 @@ class BasketComponent extends \OxidEsales\Eshop\Core\Controller\BaseController
             ];
         }
 
-        if (is_array($aProducts) && count($aProducts)) {
+        if (\is_array($aProducts) && \count($aProducts)) {
             if (Registry::getConfig()->getRequestParameter('removeBtn') !== null) {
                 //setting amount to 0 if removing article from basket
                 foreach ($aProducts as $sProductId => $aProduct) {
--- a/source/Application/Component/CategoriesComponent.php
+++ b/source/Application/Component/CategoriesComponent.php
@@ -241,7 +241,7 @@ class CategoriesComponent extends \OxidEsales\Eshop\Core\Controller\BaseControll
                 // setting list type directly
                 $sListType = 'manufacturer';
                 $sActCat = $sActManufacturer;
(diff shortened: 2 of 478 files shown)


Sieg:
Hi @bpesch, we have discussed the PR in the team, and I should say - quite a lot ;) So, what you did is great, but I would like to ask several questions to make this PR not so "painful" for us:
* Does the tool you have used have the fixer for compiled methods only? - https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048 ('opcache-only' => true param is mentioned there for CSFixer).
* Would it be possible to not actually prefix all those methods mentioned, but import them instead? Maybe even as a grouped import? (eg. use some\namespace\{ClassA, ClassB, ClassC as C};)

If those points would be possible, and we would have the PR that fits those rules, we would be really happy to merge it!

bpesch:
Hi @Sieg,

the methods affected by this are defined in this function in the php interpreter: https://github.com/php/php-src/blob/php-7.2.6/Zend/zend_compile.c#L3872 plus some additional functions that are being inlined by opcache: https://github.com/php/php-src/blob/php-7.2.6/ext/opcache/Optimizer/pass1_5.c. 
I suppose PHPCSFixer only applies the fix for the functions actually affected and the opcache inlining is the reason for the mentioned parameter.

Import the native functions using the `use` keyword has the same effect as prefixing the backslash.

Sieg:
Hey @bpesch 

* By "compiled methods" I mean the ones that affected by opcache, and we would like to have only those cases improved for now.
* use function keywords have the same effect but: will not touch so many lines in code and with our development merging strategy will not introduce us so many conflicts later; code doesn't look nice anymore for a lot of people (even if it's not nice now, it will be even more ugly).

Sieg:
Hey @bpesch feel free to reopen the PR or create a new one with a link to this one when communicated changes will be done. This one on its current state is not relevant anymore anyway.

PR #844, opened 2020-11-02
Apply modern coding standards

bpesch:
Includes improvement proposed in https://github.com/OXID-eSales/oxideshop_ce/pull/837


Sieg:
Hey @bpesch, looks like our current tools don't like your changes very much.

bpesch:
Hi @Sieg, 

this was meant as a suggestion how ECS could take over from the conventional phpcs which is implemented right now.
Adding type hints, strict comparisons and the `strict_types` declaration are most likely going to break some code thats not written properly.

This could be fixed step-by step using the set configuration.

Sieg:
Ok, thanks for showing us this. I will take this tool name with your PR id to one of our team meetings, and we will decide if there is some time for checking and applying it. For sure, it will not be merged as it is. Thanks!

PR #845, opened 2020-11-05
Resetting voucher reservations after logout

JaroslavHerber:
The reservation of a voucher is not automatically resetted after logout.
If not resetted, the voucher is not redeemable again for some time.

Vouchers are not saved in user-accounts, so they are not really "reserved" after logout. Therefore we have to reset the reservation.


Sieg:
Hey @JaroslavHerber, thanks for the improvement!  Merged to b-6.2.x and up to master.

PR #846, opened 2020-11-16
Change access to getRenderer method in Email class

pvanhemmen:
The Email class is often extended with own send methods for additional email functionality. To adapt the core principals of the logic from this class it is crucial to have access to the getRenderer method in extended classes, otherwise it needs to be reemplemented every single time.

### DIFF
--- a/source/Core/Email.php
+++ b/source/Core/Email.php
@@ -399,7 +399,7 @@ class Email extends PHPMailer
      *
      * @return TemplateRendererInterface
      */
-    private function getRenderer()
+    protected function getRenderer()
     {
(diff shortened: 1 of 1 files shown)


dx-bhesse:
Maybe getRenderer() could also get renamed to getSmartyRenderer() (or as alias-function) and an additional getTwigRenderer() would be nice

Sieg:
Hey @pvanhemmen, thanks for the fix. Merged.

@dx-bhesse, thanks for your interest! I believe, there will be no specific template engine type methods at the end, and the renderer will just come as a dependency and be implemented by following some renderer interface.

PR #847, opened 2020-11-27
remove display errors 0

keywan-ghadami-oxid:
ini_set('display_errors', '0') does hide errors from developers so it should not be done in the bootstrap.
I know it is important security setting and should be set to 0 in production for web requests, but this is a setting the hoster is responsible to do, and the application should not change it from my point of view.
If this setting is overwritten here it must be configurable and distinguish at least between CLI and web request, wich would be a overklill here and that is what php.ini is good for.


Sieg:
Hey guys, #728 was merged recently. Thanks.

PR #848, opened 2020-12-07
OXDEV-4178 Create basket private by default

NikolaIvanovski:


AlfonsMartin:
Seems a valid point for me, as basket should not be public by default.

PR #851, opened 2020-12-26
prevent password autofill

vanilla-thunder:
the "new password" field in user management triggers browser autofill function and might overwrite the password of selected user if you do not clean this field manually.
Adding `autocomplete="new-password"` to this field will prevent browser's autofill function


Sieg:
How about autocomplete="off" ?

Sieg:
Ok, I have double-checked several discussions around and looks like your solution is the correct one! Thanks for the fix! Merged.

PR #854, opened 2021-02-18
feature: get pictures from first variant in MdVariants without images in masters

SvenBrunk:
This is used in one of our OXID cloud projects to get the pictures of master products in product overview pages from the first active variant instead.
In this case the ERP system does not have real products and also no images assigned to them for master products.
With these modifications it can be used e.g. in listitem_grid.tpl by adding [{assign var="oPictureProduct" value=$oView->getPicturesProduct()}] and replacing $product->getThumbnailUrl() with $oPictureProduct->getThumbnailUrl()

### DIFF
--- a/source/Application/Component/Widget/ArticleBox.php
+++ b/source/Application/Component/Widget/ArticleBox.php
@@ -113,6 +113,26 @@ class ArticleBox extends \OxidEsales\Eshop\Application\Component\Widget\WidgetCo
     }
 
     /**
+     * Returns pictures product object
+     *
+     * @return Article
+     */
+    public function getPicturesProduct()
+    {
+        /** @var Article $oProduct */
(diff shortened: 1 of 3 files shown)


SvenBrunk:
I marked this pull-request as draft as it is not meant to go into the product like this. This should be a configurable option and work for all kinds of non-buyable parents with all kinds of variants (not only md variants)

Sieg:
Hey @SvenBrunk, looks like a performance problem, no?

Sieg:
Variant is getting the images from master, if it does not have its own images. It can be even merged together, like - first image can be coming from master, and all other from the variant if needed. Loading the first variant is much more complicated as it requires configuration for main variant, and not sure if its really needed, as master can be reused and its already done.

PR #855, opened 2021-02-18
Autoselect single variants

SvenBrunk:
This is the code I could find for the code/core part of autoselecting variant dimensions that have no possible choice left.

Not intended to be merged into the product as is. There are also other modules within OXID Professional Services that try to achieve similar things.


Sieg:
Added to Backlog refinement.

PR #856, opened 2021-03-05
Improve redirect cycle detection.

keywan-ghadami-oxid:
This fixes issues when redirect cascade is happening.

## How to Reproduce:

The old code successfully stops the browser from running in a cycle, but it is not able to detect if it is really a cylce or just a second redirect after the first one.
In the project where I hit this limitation there was a simple redirect from basket to startpage if the user is not logged in.
As the normal logout function (in ssl mode) does a redirect, a logout on the basket page caused a sequenze of two redirects.

## Redirect Counting

Counting the redirect does not really detect a cycle but is good enough to detect problems. The hardcoded 3 is arguable some browser may allow 16 or more.
Having a serverside cycle detection is better to be able to provide better error messages and to be able monitor the application for errors on server side, the value should therefore be lower then 16.
Also multiple redirects should be avoided to have good user experience so value lower then 4 is preferable but also you do not want to optimize big parts of the code base so 1-3 redirects may be acceptable
see
https://stackoverflow.com/questions/10895406/what-is-the-maximum-number-of-http-redirections-allowed-by-all-major-browsers

## Use of Exception Instead of Return

The old code used a simple return statement, so the caller code could continue handling the edgecase but after research I found no caller code in projects and module that was able to handle this return correctly. often the caller code assumed that the redirect function will never return, so the behavior was causing following errors that were hard to understand.
It also not a good alternative to exit the code if a cycle is detected because the user needs some feedback in the browser to understand that a error did happen and to navigate back to start page or to report the issue.
Throwing a exceptions allows the caller code to deal with the exception and also to framework to take actions in that situation (e.g. showing the offline page and logging the error) which helps developers to understand and handle the situation.

### DIFF
--- a/source/Core/Utils.php
+++ b/source/Core/Utils.php
@@ -1007,7 +1007,7 @@ class Utils extends \OxidEsales\Eshop\Core\Base
         //#M341 and check only if redirect parameter must be added
         $redirectCount = \OxidEsales\Eshop\Core\Registry::getRequest()->getRequestParameter('redirected', 0);
         if ($blAddRedirectParam && $redirectCount > 3) {
-            throw new RedirectException("to manny redirects");
+            throw new RedirectException("Too many previous redirects were detected when forwarding to $url. OXID eShop allows a maximum of 3 redirects to ensure a good user experience");
         }
 
         if ($blAddRedirectParam) {
--- a/source/Core/Utils.php
+++ b/source/Core/Utils.php
@@ -998,8 +998,8 @@ class Utils extends \OxidEsales\Eshop\Core\Base
      * @param string $sUrl               URL to be redirected
      * @param bool   $blAddRedirectParam add "redirect" param
      * @param int    $iHeaderCode        header code, default 302
-     * @throws RedirectException
-     * @return no-return
+     * @throws RedirectException When redirect cycle is detected
+     * @return no-return Method does never return
      */
     public function redirect($sUrl, $blAddRedirectParam = true, $iHeaderCode = 302)
     {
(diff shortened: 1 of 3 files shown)


source/Core/Utils.php
```diff
@@ -997,19 +998,20 @@ public function showOfflinePage()
      * @param string $sUrl               URL to be redirected
      * @param bool   $blAddRedirectParam add "redirect" param
      * @param int    $iHeaderCode        header code, default 302
-     *
-     * @return null or exit
+     * @throws RedirectException
+     * @return no-return
```
flow-control, line 1002: I understand what you mean by `no-return`, but I would suggest using `void` and express the nature of this method in the description of the `@return` annotation. What do you think?

```suggestion
     * @throws RedirectException When redirect cycle is detected
     * @return void Method does not return in case redirect can be executed
```

keywan-ghadami-oxid: "no-return" can be used by static code analysers  
and is supported by psalm, phpstan and others:
- https://psalm.dev/docs/annotating_code/type_syntax/atomic_types/
- https://phpstan.org/writing-php-code/phpdoc-types
- phpstorm
and there is open pull request to get it into php-fig

https://github.com/php-fig/fig-standards/pull/1191

I understand that it is still not standardised, but it helps to find issues when developing.

flow-control: TIL :wink: 
I checked the same page in the PHPStan docs, but for some reason I couldn't find it at my first try, now it worked. In this case I would also omit the description :wink:

alfredbez: Hi guys, just wanted to let you know, that Matt and Ondrej are working on a RFC to get this into the PHP Core: https://wiki.php.net/rfc/noreturn_type

&lt;/OT&gt;

Sieg:
Hi @keywan-ghadami-oxid,

Could you please describe / provide some more examples of those broken cases? Maybe some snippet/module so we can play with and see the problem ourselves? This would drastically change the time needed for making decisions like this.

PR #857, opened 2021-03-09
Save Remarks loses oxcreate date

SeifertDaniel:
https://bugs.oxid-esales.com/view.php?id=7217

### DIFF
--- a/source/Application/Model/Remark.php
+++ b/source/Application/Model/Remark.php
@@ -50,9 +50,9 @@ class Remark extends \OxidEsales\Eshop\Core\Model\BaseModel
     {
         if ($blRet = parent::load($oxID)) {
             // convert date's to international format
-			$this->assign([
-				'oxcreate'	=> \OxidEsales\Eshop\Core\Registry::getUtilsDate()->formatDBDate($this->oxremark__oxcreate->value)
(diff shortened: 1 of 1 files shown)


Sieg:
Hi Daniel!

I am not able to reproduce the issue. Could you tell me more about your installation? Do you have some special format for your database? Are you using our oxvm vagrant machine for the test?

SeifertDaniel:
Hi Sieg.

I was able to reproduce the behaviour in an EE 6.2.3 (without modules, minimal database changes) in the OXVM. This shouldn't really matter, since the problematic code runs independently of the system. I have documented the steps with screenshots.

The same entry is always highlighted in the Admin screenshots. Please also note the sort order of the entries in the admin area after saving.

The bug not only changes the oxcreate date, but also all types that are not "r".

1 1 database before saving
1 2 admin panel before saving
2 1 database after saving
2 2 admin panel after saving

Sieg:
Hey @SeifertDaniel, thanks for more information. So the case was reproducible if the German language is selected as admin language during the login screen, not actual date formats changing in shop settings.

Merged the fix to b-6.2.x and up. Thanks! Have a nice day!

PR #859, opened 2021-03-15
Provide getConfigParam() in $oViewConf/Smarty

JaroslavHerber:
Currently you have to assign oConfig separately to get a config-param in templates:
[{assign var="oConfig" value=$oViewConf->getConfig()}]
[{$oConfig->getConfigParam('sDefaultLang')}]

Smarter way would be:
[{$oViewConf->getConfigParam('sDefaultLang')}]

Alternative way:
Make $oConfig-var global in templates like $oView or $oViewConf vars

### DIFF
--- a/source/Core/ViewConfig.php
+++ b/source/Core/ViewConfig.php
@@ -1258,6 +1258,19 @@ class ViewConfig extends \OxidEsales\Eshop\Core\Base
 
 
     /**
+     * return config-param value
+     *
+     * @param string $sName param name
+     *
+     * @return mixed
+     */
+    public function getConfigParam($sName)
+    {
+        return Registry::getConfig()->getConfigParam($sName);
+    }
+
+
+    /**
      * Returns true if selection lists must be displayed in details page
      *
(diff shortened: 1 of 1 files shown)


vilmal:
Hi @JaroslavHerber, thanks for your interest in improving the product!

Actually, the ``getConfig`` method is deprecated for a while and will be removed completely in the future. I would strongly advice you not using it. 

The main idea is that you should access only theme specific settings in the template, for that please use:

```
$oViewConf->getViewThemeParam('someThemeConfigParameter');

```

We do not recommend to access other config parameters in the template. But if you have to do it, extend your controller accordingly and use the ``ViewConfig::setViewConfigParam('someVar', 'some Value')`` method to send required config parameter from your controller to your template:

```
$this->getViewConfig()->setViewConfigParam('myParam', 'myValue');
```

Once you’ve used it, the config parameter can be accessed in your template: 

```
$oViewConf->getViewConfigParam('someVar');

```

Best Regards,
Vilma

PR #860, opened 2021-03-16
Missing Event SettingChanged

TumTum:
As the [documentation](https://docs.oxid-esales.com/developer/en/6.2/development/tell_me_about/event/ModuleEvents/SettingChangedEvent.html) says, the event should be triggered when the module settings change. Therefore I expect the event also when I change module settings in admin.

### DIFF
--- a/source/Internal/Framework/Module/Setting/SettingDao.php
+++ b/source/Internal/Framework/Module/Setting/SettingDao.php
@@ -10,11 +10,13 @@ declare(strict_types=1);
 namespace OxidEsales\EshopCommunity\Internal\Framework\Module\Setting;
 
 use OxidEsales\EshopCommunity\Internal\Framework\Config\Utility\ShopSettingEncoderInterface;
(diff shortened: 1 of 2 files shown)


Sieg:
Hi @TumTum,

So this event is already triggered in ModuleSettingBridge: https://github.com/OXID-eSales/oxideshop_ce/blob/1c7ecf404dea372a30969d897d75aed10e77dbf2/source/Internal/Framework/Module/Configuration/Bridge/ModuleSettingBridge.php#L65

Am i missing something? :/

TumTum:
Hi @Sieg,
yes true ... but this way is not used when you save a module configuration in the admin backend. No event is triggered because it does not use the ModuleSettingBridge.

flying-hirsch:
Hi @TumTum , thank you for this PR!

SettingDao looks like a correct place to dispatch this event. I think we should also remove it from the  ModuleSettingBridge to do not dispatch it twice in some cases.

TumTum:
Ah Okay, what should I do with the pull request now?

Sieg:
Hey @TumTum, thanks for the fix!

PR #863, opened 2021-03-26
fix endless loop on module error

keywan-ghadami-oxid:
in rare case a module overloads \OxidEsales\Eshop\Core\Module\Module and gets for example deleted 
then the shops must not try to call oxNew for \OxidEsales\Eshop\Core\Module\Module because that would create a endless loop.

this kind of loops can result very hard to debug segmentation fault (because out of memory errors in combination with xdebug)

My first try to fix this was not sufficient, anyway it looks like this regression is only in oxid eShop 6.2 but not before, at least I did not hit it in oxid 6.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [6.8.1] - Unreleased
+
+### Fixed
+- Use not extended OxidEsales\Eshop\Core\Module\Module in module chain generator error case. [PR-863](https://github.com/OXID-eSales/oxideshop_ce/pull/863)
+
 ## [6.8.0] - 2021-04-13
 
 ### Added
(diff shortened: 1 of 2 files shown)


liulka-oxid:
If the following reproduction steps are correct:
- module makes an extension for the class, annotated as `@internal Do not make a module extension for this class.`
- module installation is getting damaged (files are improperly removed?)
then I think we should consider other solution (without changing instantiation method in Core class)

keywan-ghadami-oxid:
yes, why? It's like someone press the red button with the tile "do not press it it" and it destroys the house of other people, then I like to still think about if that button really needs to do so much damage.

The use-case here is to remove old modules, (e.g. while upgrading OXID eShop). Usually the dev that faces this kind of issues is not to blame (so I may be to blame to be involved in illegal extension of that "internal" namespace, but usually other devs suffer from this errors). 

There is already a extra check for the `Config` class in another location, also by using new instead of "oxNew" but it does not cover the `Module` class yet. Using `new` instead of `oxNew` is explicitly telling that you do want the instance without module extensions for this class.

Another way of dealing with the issue would be to filter bad/forbidden extensions. In this case you can not do it only in the validation process of the metadata because that configuration may be already in the database. So I am thinking about using `new` instead of `oxNew` for all internal classes should also be valid.

Sieg:
Hey @keywan-ghadami-oxid, I have just tried to reproduce the issue on one of currently installed local modules:
* Extended the module class in my special module.
* Registered, installed and activated the module.
* Deleted the class first, shop is not crashing, it shows the file in problematic files list in admin.
* Then i have deleted the whole module without unregistering or deactivating it - still everything is working, just module is marked as problematic, with all files listed:
image

I am not sure what you are doing there with those modules, so it going for infinite loops for you ;) Could you please double-check the case and write more instructions on reproducing this?

dx-bhesse:
https://github.com/OXID-eSales/oxideshop_ce/pull/863 is probably related.

Sieg:
Hey guys, thanks for the feedback here. As the Module class is marked as @internal by us, it's a good idea to use this as proposed by Keywan in this case, so we will ensure proper calculation in the case of error.

PR #864, opened 2021-04-21
replace copied and pasted comment from DeliveryController with real info

SvenBrunk:


Sieg:
Hey @SvenBrunk, thanks for the fix.

PR #867, opened 2021-05-07
Fix typo in ActionList.php

matasskarzauskas:
Fixed typo in SQL statement in ActionList.php. 

Previously this SQL query would throw fatal error:
`Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'oxtypex' in 'where clause' `


Sieg:
Hello @matasskarzauskas, thanks for the fix!

PR #869, opened 2021-05-28
The bug occurs when having the B2B modules activated.

BernhardScheffold:
When increasing the amount of a basket position in the basket overview so that a delivery rule which has zero cost
should be pulled, the old cost is still visible and only updated when pushing the update button a second time.
Initializing the member variable _aDeliveries to an emty array fixes this behaviour.


Sieg:
Hey @BernhardScheffold, looks like this one is merged already (b-6.3.x), and i forgot to close it. Have a nice day! :)

PR #870, opened 2021-06-08
Count assigned variants for category article count

kingschnulli:
Currently assigned variants are ignored when calculating number of category articles, this will hide categories only containing variants when using "blVariantsSelection" option.


Sieg:
Hey @kingschnulli, thanks for the pull request!

I have tested this for a moment, and I see it cannot be merged as is - in your case, we will get variants counted with other category products, but variants themselves are not shown in category, so you will result with count 4, but 2 products inside for example.

Feel free to reopen the PR if you have more information I could miss. Thanks.

PR #873, opened 2021-06-25
feat: style of password-input-fields

mariolorenz:
in the same style as text-input-fields


mariolorenz:
https://bugs.oxid-esales.com/view.php?id=7249

BenjaminJoerger:
Travis breaks because global getViewName was ditched recently. Please use the according getViewName method.

Sieg:
Hey @mariolorenz, thanks for the fix. Rebased on b-6.3.x and merged up.

PR #874, opened 2021-06-28
fix: PasswordHandling in Admin Tpls

mariolorenz:
since Update of backend jquery (from v1.5.1 to v3.4.1 :
OXDEV-2401 OXDEV-2501 Update to latest jquery lib)
the password-validation is broken because of using "attr"
instead of "prop" (introduced in jquery v1.6)

also the data-attributes should written in lowercase letters


mariolorenz:
https://bugs.oxid-esales.com/view.php?id=7249

Sieg:
Hey @mariolorenz, thanks for the fix. Rebased on b-6.3.x and merged up.

PR #877, opened 2021-09-15
Changed forge url to official docs

KiryiMONZTA:
The system requirements URL was linking to an OXIDforge article which is not available anymore. I changed this to the official docs. Since the structure of the official docs are slighty different, the anchors needed to be adapted as well. This also introduced a second URL + anchors due to split pages for this information (requirements + preparation).


Sieg:
Hello @KiryiMONZTA, thanks for fixing this.

PR #878, opened 2021-09-28
Remove redirects to directory

keywan-ghadami-oxid:
why:
It can make trouble if apache adds internal ports to the redirect, it makes oxid not flexible to configure seo url by modules (e.g. you may not like to have the slash or .html extension and it does not add value because seo.php can also handle the redirect so it is useless code.

bc break: no

performance:
Some redirects will get slower because they will now be handled by php, normal performance of a shoul should stay the same as this redirects are not used when normally using the page, but only when typing urls manually.

### DIFF
--- a/source/.htaccess
+++ b/source/.htaccess
@@ -21,11 +21,6 @@
     RewriteCond %{REQUEST_URI} !(\/admin\/|\/Core\/|\/Application\/|\/export\/|\/modules\/|\/out\/|\/Setup\/|\/tmp\/|\/views\/)
     RewriteCond %{REQUEST_FILENAME} !-f
     RewriteCond %{REQUEST_FILENAME} !-d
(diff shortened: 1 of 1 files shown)


keywan-ghadami-oxid:
I saw the need for this in multiple projects already, a the default makes it complicate to create generic modules and solutions for this. 
About performance: yes php is slower then mod rewrite in handling the redirects, but usually this redirects handled by modrewrite are not the normal use case, so for a low amount of redirects it doesn't matter to add some milliseconds overhead - on the other hand it brings great flexibility and reduces the configuration complexity and will allow generic OXID SEO modules.

This is still a working draft, any feedback is welcome.

Sieg:
Hi @keywan-ghadami-oxid,

Could you please describe / provide some more examples of those broken cases? Maybe some snippet/module so we can play with and see the problem ourselves? This would drastically change the time needed for making decisions like this.

PR #879, opened 2021-10-01
Fix db fetchmode in seoencoder loadfromdb function

C3000:
How to Reproduce:
In the seoencoder in the function loadFromDb there is a bug with the fetchMode.
Th database is retrieved by the DatabaseProvider with the fetchMode ASSOC.
But in the function loadFromCache the fetchMode is switched by generating a view id (especially by detemining the sorting with a qoute)
So the retrieve of the database must be placed beneath this function.
Otherwise the seourl could never be found because there is never a key called oxseourl in the fields.


michaelkeiluweit:
Similar (resolved) issue: https://bugs.oxid-esales.com/view.php?id=6892

PR #881, opened 2021-10-15
translations are now working for both, frontend and backend

moritzdemmer:
### why
When also using the frontend translations for the backend, no extra backend translation files are needed for modules which makes it less error-prone. It's also possible to use standard Oxid Frontend translations in the backend without copying the keys.


Sieg:
Hi @moritzdemmer, thanks for trying to make working with translation easier!

Actually, this question appeared several times already during the last years, and even though the fix looks simple, the one like this will introduce the undocumented and sometimes unwanted inheritance of translations, which we cannot allow to be merged at this point in time.

I believe, in the future, we will have more possibilities of configuring your translation files lists for your modules, and it will be properly tested and documented.

PR #882, opened 2021-10-15
Allows to throw own exception messages in admin login

moritzdemmer:
### Why
There is no possibility to throw your own error messages in the admin login area as the message is hard coded.
When adding new functionality to the admin login, it may be necessary to show own messages instead of the "Wrong Username/Password",


Sieg:
Hello @moritzdemmer. Thanks for the improvement. Merged to b-6.4.x and up to master with several changes.

PR #885, opened 2021-11-19
Fix return type in phpdoc

keywan-ghadami-oxid:
return type must include null because it multiple places do return without return without value


Sieg:
Hey @keywan-ghadami-oxid, thanks for the fix.

PR #888, opened 2022-02-07
The security has nothing to do with the domain, but with connection type.

TumTum:
The bug is especially a problem in the EE version.
In the admin backend, no session is saved when you are in https mode and make a shop change where mall URLs are stored.
Moreover, it is illogical the security does not depend on the domain but on the connection. In http mode everything works fine and https mode does not. The cookie is issued anyway. It is just not returned at a https connection.
If no session is saved in https there are many language problems in the admin-backend. See https://bugs.oxid-esales.com/view.php?id=7306


Sieg:
Hey @TumTum, thanks a lot for the fix!

PR #889, opened 2022-02-11
extend handling for service response errors

MaxBuhe01:
Why?
- missing handling for service response errors

What did we do?
- add error cases that lead to a new request


Sieg:
Hey @MaxBuhe01, thanks for the improvement!

PR #891, opened 2022-04-11
Fixed rare exception if getPrice() is null

JaroslavHerber:
In some extremely rare circumstances, $oBasketItem->getPrice() can return null. Then, line 234 will throw a PHP exception because getPrice() cannot be called on a null-object. (I could not really reproduce the null-object by myself - but somehow one customer produced this.)
There is already same check in line 255, which I added to line 233 as well.

### DIFF
--- a/source/Application/Model/Delivery.php
+++ b/source/Application/Model/Delivery.php
@@ -230,8 +230,8 @@ class Delivery extends \OxidEsales\Eshop\Core\Model\MultiLanguageModel
                 case self::CONDITION_TYPE_PRICE: // price
                     if ($this->getCalculationRule() == self::CALCULATION_RULE_FOR_EACH_PRODUCT) {
                         $dAmount += $oProduct->getPrice()->getPrice();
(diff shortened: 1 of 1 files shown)


JaroslavHerber:
UPDATE: the null-price-object can be producted by adding a product to basket via code without recalculating the basket afterwards:
1. Registry::getSession()->getBasket()->addToBasket(...);
2. // Do not execute...! -> Registry::getSession()->getBasket()->calculateBasket(true);
3. The price-object is missing in the basket-session for the added basketitem
4. If you use Delivery->getDeliveryAmount(...) with this basketitem, you get an exception, because the price-object was not calculated previously (step 2)

Sieg:
Hi @JaroslavHerber, thanks for the fix, looks good!

I have rebased on b-6.5.x and merged at https://github.com/OXID-eSales/oxideshop_ce/commit/20306c15c14edece590c993c67db8943cd6718a0

PR #893, opened 2022-05-04
Change type of default value for parameter (probably missed before)

SvenBrunk:
ERROR: InvalidParamDefault - project-modules/bergspezl/Model/Article.php:275:15 - Default value type string() for argument 1 of method OxidProfessionalServices\Bergspezl\Model\Article::getZoomPictureUrl does not match the given type int (see https://psalm.dev/062)


Sieg:
Hey @SvenBrunk, thanks for the fix.

PR #894, opened 2022-05-04
Fix used logging mechanism in shutdown callback

d3mo17dot:
Enable to handle uncaught errors with custom defined logger (monolog)


Sieg:
Hey @d3mo17dot, thanks a lot for the Pull Request! As I remember, this place was handled like this by intention, to be more bullet-proof. The getLogger may cause another chain of loads that may explode as well.

Let's double-check with architects.

Sieg:
Let's not change that for now. The reason mentioned above. Thanks!

PR #895, opened 2022-05-05
perf optimization payment list.

keywan-ghadami-oxid:
question in slack by aurimas urbonas. would it be faster with this small change without any break?


abejingas (approved):

abejingas:
At least with the setup of my client's database (with a lot of records in oxobject2group) there is a dramatical change without this join

Sieg:
Looks good. Thanks for the improvement!

PR #897, opened 2022-06-30
Fixed PHP-warning if varselid is not an array

JaroslavHerber:
Search-robots like to add this param to detail-pages they crawl: ...?varselid=Array ("Array" is a string here)
And this will cause a PHP warning, if we don't check the param with is_array()


Sieg:
Hi @JaroslavHerber, could you please sign the updated CLA?

PR #900, opened 2022-09-09
Remove dynamically included Google Webfonts (GDPR) from offline.html

SvenBrunk:
Removed inclusion of Google Webfonts directly from Google and replaced it by multiple alternatives: 1) still Source Sans Pro in case the browser has it from any other source 2) OXID font in case anyone uses it
3) Helvetica and Arial for 2 of the most widely used fonts installed on most devices 4) generic sans-serif (as before)


Sieg:
Hi @SvenBrunk, please sign the updated CLA.

PR #901, opened 2022-10-19
remove google fonts

proudcommerce:


Sieg:
Hi @proudcommerce, looks like a copy of #900 ? In any case, Thanks for the effort!

PR #902, opened 2022-10-24
Various changes to decrease amount of PHP8 warnings

geeky:
This adds various casts and types in function signatures to lower the amount of php8 specific warnings.
Also changes some instances of null default values to using an empty string instead.
Removed "attribute"-mapping from BackwardsCompatibilityClassMap.php, as php8 itself introduced an attribute-class.
Changes also that users are only logged out when an error occurs (in bootstrap.php) if the shop is not running in debug mode

### DIFF
--- a/source/Application/Component/BasketComponent.php
+++ b/source/Application/Component/BasketComponent.php
@@ -272,19 +272,19 @@ class BasketComponent extends \OxidEsales\Eshop\Core\Controller\BaseController
     {
 
         // active controller id
-        $controllerId = Registry::getConfig()->getRequestControllerId();
+        $controllerId = (string)Registry::getConfig()->getRequestControllerId();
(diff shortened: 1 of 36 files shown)


Sieg:
Hey @geeky, thanks for the effort on this Pull request, sadly there are several issues with it:

* its too big - too many different questions are involved, and we cannot merge all of it for sure
* some of the changes are changing the way methods have been working, doing something like that is a complicated process with a deprecation procedure
* what code does is more important than what is stated in the docblock, so don't try changing methods to fit docblock, but change the docblock to fit the method
* changing the method descriptions (like adding typehints for parameters) breaks backward compatibility as well, again, deprecation process is required

Consider making small pull requests for fixing specific problems / improve specific questions. Smaller it is (the PR) - bigger chance of having it merged.

Regards, Anton.

PR #903, opened 2022-12-16
change oxremark oxtype field type from enum list to varchar

SeifertDaniel:
to also allow entries outside the predefined list


SeifertDaniel:
Enum is unfavourable for extensibility because it does not know a command to simply insert an additional value type. The existing value set must only be determined and completed and then stored in the structure via time-consuming detours. 

However, the enum type can store and process different values more efficiently due to its storage as an indexed list. 

We have been able to find a technical solution for the disadvantage with the extension. Therefore, I close the pull request in view of the higher efficiency.

Sieg:
Hey, @SeifertDaniel thanks for the information. The pull request in its current state would be not enough anyway - all the database changes come through migrations written in a way that ensures data integrity.

Could you briefly describe how you manage to solve the problem with additional remark types?

PR #906, opened 2022-12-29
Optimize accessing 'oxorder__oxbillcountry'

geeky:
(Accessed from e.g. order_owner.tpl)

The__get()-call for 'oxorder__oxbillcountry' calls getBillCountry(), but getBillCountry() again accesses oxorder__oxbillcountry with: if (!$this->oxorder__oxbillcountry->value) {
which creates an 'undefined property' warning (would create an endless loop if php did not have special handling for magic getters)

This patch introduces $this->_sBillCountryTitle for use in getBillCountry() and does no longer check $this->oxorder__oxbillcountry->value. And __get('oxorder__oxbillcountry') now always calls getBillCountry(). 
The same applies for 'oxorder__oxdelcountry'

### DIFF
--- a/source/Application/Model/Order.php
+++ b/source/Application/Model/Order.php
@@ -235,21 +235,7 @@ class Order extends \OxidEsales\Eshop\Core\Model\BaseModel
      * @var \OxidEsales\Eshop\Application\Model\Basket
      */
     protected $_oOrderBasket = null;
-
-    /**
-     * Billing country name from billing country id
-     * (Call getBillCountry())
-     * @var null|string
-     */
-    protected $_sBillCountryTitle=null;
-
-    /**
-     * Delivery country name from delivery country id
-     * (Call getDelCountry())
-     * @var null|string
-     */
-    protected $_sDelCountryTitle=null;
-
+    
     /**
      * Class constructor, initiates parent constructor (parent::oxBase()).
      */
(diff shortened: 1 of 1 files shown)


Sieg:
Hey @geeky, thanks for your interest in improving the product.

For accessing model data, try using ``$model->getFieldData('oxbillcountry')`` method (and getRawFieldData if needed), which checks for object existence inside and should not produce notices/warnings.

Also, we cannot ignore the current "oxbillcountry" field value, somehow, its value should be respected too.

Another point for you to check - is the quality requirements for the new code:
* it should be PSR 12 compatible (https://docs.oxid-esales.com/developer/en/latest/development/modules_components_themes/quality.html#code-quality-requirements)

Please check my points, and let me know if it helps to deal with issues.

geeky:
> For accessing model data, try using `$model->getFieldData('oxbillcountry')` method (and getRawFieldData if needed), which checks for object existence inside and should not produce notices/warnings.
That actually does produce the same warning:
getFieldData('oxbillcountry') calls getFieldIfAvailable(), which then checks via:
return ($this->$longFieldName instanceof Field) ? $this->$longFieldName : null;
but "$this->$longFieldName instanceof Field" also causes the "Undefned property" warning, as the access to $this->$longFieldName also triggers __get('oxorder__oxbillcountry') again.
For a property that exists BaseModels __get() would normally initialize that field on-the-fly then, but thats never called for 'oxorder__oxbillcountry' as the Order-Model itself handles that property in its __get().

('oxorder__oxbillcountry' / 'oxbillcountry' does not actually exist in the oxorder database table, its only a 'virtual'-field, so BaseModel's initDataStructure() wills also never initialize that field. Only 'oxbillcountryid' actually exists)

I have now refactored it so that getBillCountry() will check for the existance of oxorder__oxbillcountry via property_exists() (which does not call __get(), which should not be a problem here) and sets the 'oxorder__oxbillcountry'-property.

Sieg:
To not make things more complicated, i decided to merge part of your solution - https://github.com/OXID-eSales/oxideshop_ce/commit/c82fbcbe37fcac0a98961a843d9397ee845789dc 

Hope that will be enough for this case. Thanks for digging up and clarifying all the questions.

PR #908, opened 2023-01-05
Remove redundant methods in UserPayment model

Sieg:


Sieg:
This PR is not actual anymore, as those methods have been removed already.

PR #909, opened 2023-01-05
Handle multidimentional arrays with Yaml

Sieg:


Sieg:
This solution needs a lot of improvements on topics:
* old data migration process?
* how to handle new format, maybe it should be converted to the old one for the display?
* new possible Exceptions not handled in any way.

PR #913, opened 2023-08-12
reset oxarticles__oxsoldamount

tabsl:


Sieg:
Hey @tabsl, thanks for the fix.

Some more context: this counter value is used in the Top articles list calculation.

PR #918, opened 2023-09-07
$this->_oUserData not existing

bYemma:
The correct naming is "_aUserData", because variable is an array.
Property "_oUserData" doesn't exist. Property "_aUserData" is defined for this: `protected $_aUserData = null;`

Stops PHP warning: "Undefined property"


michaelkeiluweit:
https://bugs.oxid-esales.com/view.php?id=7533

PR #936, opened 2024-01-30
Fix Discount Calculation if an Article is assigned to Coupon

janspeller:
The Calculation of the Discount Value is not precise if:
- Coupon has an Article assigned
- Article Quantity is higher than 1
- Article has odd price (like 1.62 used for the written Test)

The Coupon Value in the Basket will be rounded to 2 Decimals before being multiplied by the Article Quantity. This leads to an inprecise Calculation.


Sieg:
Hey @janspeller, is the test you added green for you?

Sieg:
@janspeller The issue here is how the shop calculates discounts based on your current currency precision settings. Let me explain.
 
For a general 10% coupon, it correctly deducts 10% (e.g., 1.62). But in your case, the discount is applied to each item individually. 
This is important because the storage application might also show each item as 10% discounted, rounding the price for each item. This is where the precision issue arises.
 
If we apply the discount as you suggest, it would treat it as a general coupon, which could cause problems with the accounting and lead to mismatched totals.
 
I agree there’s an issue. We can't currently separate calculation and display settings to make your requested calculation work without affecting the final price display.
 
To move forward, please go to Master Settings -> Core Settings -> Settings -> Other Settings, find the Currency configuration, change the precision for EUR to 3 decimal places, and see how the discount is calculated.

Sieg:
Closing this topic for now, as we cannot take the PR as you suggested, but there is a plan to improve this part in the future, so it will be more configurable. Anyway, thank you for your interest and trying to improve this!

PR #944, opened 2024-02-27
Tweak: Added a check if the user already exists for the "oe:admin:c…

oxidMichaelWalter:
Tweak: Added a check if the user already exists for the "oe:admin:create" command

### DIFF
--- a/source/Internal/Domain/Admin/Dao/AdminDao.php
+++ b/source/Internal/Domain/Admin/Dao/AdminDao.php
@@ -44,7 +44,7 @@ class AdminDao implements AdminDaoInterface
     }
     
     # check if username for email and shop id exists
-    public function userNameExits(string $email, int $shopId): bool {
+    public function userNameExists(string $email, int $shopId): bool {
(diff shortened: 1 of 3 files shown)


Sieg:
Hey, @oxidMichaelWalter, need some rework here:
* In Dao, if we could be not so specific, we could use the method for more cases, how about making a method with the name getAdminUserByEmail or even better getByEmail as it already have "create"?
* Instead of checking for user existence in the AdminFactory (and EXIT? you cannot use exit), let's check for user existence in the command itself, near email validation maybe, and we could use getByEmail for this check.
* Also, where are the tests? ;) will not work for this case.

RahatHameed:
Hi @oxidMichaelWalter, 

[Branch](https://github.com/OXID-eSales/oxideshop_ce/compare/b-7.4.x...b-7.4.x-pr944) is now waiting for review at oxid repository, will proceed further once it is reviewed.

Sieg:
Hey @oxidMichaelWalter, so this one was merged some time ago at 4efba392fc3896339d2a3b2341248734c2215b01, Thanks.

PR #962, opened 2024-04-27
can't get EOF property from integer

SeifertDaniel:
By the way: Where are your model test scripts?


Sieg:
Hello @SeifertDaniel, regarding tests for this part, can be found here: https://github.com/OXID-eSales/tests-deprecated-ce/tree/b-7.0.x

Sieg:
Hello @SeifertDaniel, thanks for the pull request. I have returned the "parent" result (where actual group removal is done) currently, It looks like better solution. Merged to b-7.1.x.

PR #963, opened 2024-05-03
can't load translation files from custom theme

SeifertDaniel:
... because of missing slash in file path


Sieg:
Hey @SeifertDaniel, thanks for the fix.

PR #971, opened 2024-06-05
Replace doc-comments with attributes to fix PHPUnit's deprecation messages

michaelkeiluweit:
Please note that I did not rebase after pushing the changes of https://github.com/OXID-eSales/oxideshop_ce/pull/970. So I had to revert that commit for this pull request. Therefore this pull request contains 3 instead of 1 commit.

Fixed messages (message too long, therefore cut down to at least deliver the idea)
```
1) Metadata found in doc-comment for method Integration\Application\Controller\Admin\ManufacturerPictureTest::testRender(). Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.

2) Metadata found in doc-comment for class Integration\Application\Controller\Admin\ManufacturerPictureTest. Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.

3) Metadata found in doc-comment for method Integration\Application\Controller\Admin\ManufacturerPictureTest::testSaveShouldThrowAnExceptionInDemoShopMode(). Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.

[...]

283) Metadata found in doc-comment for method OxidEsales\EshopCommunity\Tests\Integration\Core\SessionTest::testSidNeededForDifferentUrls(). Metadata in doc-comments is deprecated and will no longer be supported in PHPUnit 12. Update your test code to use attributes instead.
```

### DIFF
--- a/tests/Unit/Internal/Domain/Contact/Form/ContactFormMessageBuilderTest.php
+++ b/tests/Unit/Internal/Domain/Contact/Form/ContactFormMessageBuilderTest.php
@@ -13,12 +13,11 @@ use OxidEsales\EshopCommunity\Internal\Transition\Adapter\ShopAdapterInterface;
 use OxidEsales\EshopCommunity\Internal\Framework\Form\Form;
 use OxidEsales\EshopCommunity\Internal\Framework\Form\FormField;
 use OxidEsales\EshopCommunity\Internal\Domain\Contact\Form\ContactFormMessageBuilder;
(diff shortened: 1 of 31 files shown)


Sieg:
Hi @michaelkeiluweit, we cannot merge this, as the version of PHPUnit we have is 10, and some of the attributes are not yet available here. If you could double-check and fix this?

PR #983, opened 2024-07-03
Essential part of https://bugs.oxid-esales.com/view.php?id=7100 as PR

SvenBrunk:


Sieg:
Hey @SvenBrunk, the issue with this approach is that adding this condition would prevent inactive voucher series from being selected at all. That could be fine, but there's already a validation in place elsewhere that throws an error if the voucher is no longer valid.

With this new condition, the voucher wouldn't be selected at all if the series is expired, resulting in a different error – "Invalid coupon" – instead of the current one - "Expired". I can't think of a simple solution that avoids this inconsistency for the current vouchers implementation.

I have added the test for the original (the one we are fixing) case, but it doesn't cover the validation part I have mentioned (as its different subsystem, and that part is still working correctly for the expired coupons if they are given as parameter, but they will not be anymore) and rebased for 7.2.x. 

Maybe this is all fine and we should still merge this?

SvenBrunk:
@Sieg okay, thanks. Please do **not** merge this. We need to think about a better solution then

PR #984, opened 2024-07-04
Make language string placeholders work with args = 0

bYemma:
Addresses bug 7669 (https://bugs.oxid-esales.com/view.php?id=7669)


bYemma:
`$args` is `false` if the param is omitted, so a check `!== false` is adequate.

https://github.com/OXID-eSales/oxideshop_ce/blob/e1caca5bdc758b02d0439b06dbb0e07d873df2f8/source/Internal/Transition/Adapter/TemplateLogic/TranslateFunctionLogic.php#L47

bYemma:
Closed with https://github.com/OXID-eSales/oxideshop_ce/commit/d4d3224a7f90b324770ce3541497acc54362c4d9

PR #993, opened 2024-10-01
Adjust seo decoder to handle sub folder shops

GM-Alex:
Currently it's not possible to configure a sub shop on a folder based domain, since the seo decoder can't handle this. With this change it's possible to setup sub shops with a domain like https://www.mydomain.com/subshop


Sieg:
Hey @GM-Alex, as this is not usual case, could you please tell me more what have you done with your webserver to make this possible at all? The shop installs to "source" directory, so what do we need for our main .htaccess or having as a VirtualHost to make this working? Have you changed anything in bootstrap also? I can spend time and make this working, but if you did already, please share :)

GM-Alex:
Hi @Sieg,

It was a customer request for an enterprise project. The idea was to get a sub shop based on the path, like the domain is the main brand and the sub shop folders are there sub brands.

Greetings,
Alex

PR #996, opened 2025-05-18
Allow oe-console oe:setup:shop to use an existing empty dabase

geeky:
This allows the setup to use an existing database if it is still empty.
(For cases where the database has to be created via an administration interface at the hoster, for example)


kenariosz:
It has been resolved.

PR #999, opened 2026-06-09
Deal with case where $_SERVER['HTTP_ACCEPT_LANGUAGE'] is not defined

mparienti:
There is some case where the `$_SERVER` has no 'HTTP_ACCEPT_LANGUAGE' key.

### DIFF
--- a/source/Application/Controller/Admin/LoginController.php
+++ b/source/Application/Controller/Admin/LoginController.php
@@ -216,6 +216,6 @@ class LoginController extends \OxidEsales\Eshop\Application\Controller\Admin\Adm
      */
     protected function getBrowserLanguage()
     {
-        return strtolower(substr($_SERVER['HTTP_ACCEPT_LANGUAGE'], 0, 2));
+        return @strtolower(substr($_SERVER['HTTP_ACCEPT_LANGUAGE'], 0, 2));
     }
 }


Sieg:
Hey @mparienti, thanks for your PR!. Several questiosn to you: how can i reproduce the issue? Is the problem registered in the bugtracker?

mparienti:
> Hey @mparienti, thanks for your PR!. Several questiosn to you: how can i reproduce the issue? 

To reproduce the bug, use curl on the admin page: curl is the tool I used to check that a software I just installed is up. Curl does not sent any http header per default hence the bug.

Note: it seems as a niche bug, but if you want any project to be successful, it needs to be nice with developers and developer use curl.

> Is the problem registered in the bugtracker?

No idea: there is no issue visible here (on github) and I forgot the url of the internal bugtracker.

mparienti:
> > Is the problem registered in the bugtracker?
> 
> No idea: there is no issue visible here (on github) and I forgot the url of the internal bugtracker.

I didn't find it by searching `HTTP_ACCEPT_LANGUAGE` in the bugtraker here: https://bugs.oxid-esales.com/view_all_bug_page.php

RahatHameed:
@mparienti There was few changed done locally, 
- I  rebased it to b-7.6.x as it can go into b-7.6 and later mergedup.
- I have repaced Use null-coalescing instead of @, This matches Core/Language::getBrowserLanguage() which is already there.
- Added changelog entry
- I tested it locally using a intergeation testa and it passed. (It wll not be added to this change)

I was not able to force push at yoru patch either allow it or please do the above mentioned changed then I will proceed further.
Thank you

RahatHameed:
@mparienti Change has been merged [here](https://github.com/OXID-eSales/oxideshop_ce/commit/544d96c4a88362fd3b73d8db3367e90da8f28198) with improvment mentione in above comment, Thanks for your PR.

